PR #4, opened 2023-05-03
Add a service client wrapper

ahoelscher-bdai:
This provides similar functionality as the existing action client wrapper. Specifically, it contains a node that is spun when waiting on the service, which allows a service to be called in a callback.


amessing-bdai (approved):

bdai_ros2_wrappers/bdai_ros2_wrappers/service_client.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __init__(self, node_name: str, service_type, service_name: str):
+        self._node_wrapper = NodeWrapper(node_name)
+        self._service_name = service_name
+        self._client = self._node_wrapper.node.create_client(service_type, service_name)
+
+    def call(self, request, timeout_sec=None):
+        """Calls the service and returns the result. This is safe to call from a callback."""
+        while not self._client.wait_for_service(timeout_sec=timeout_sec):
+            self._node_wrapper.node.get_logger().info(f"{self._service_name} not available, waiting again...")
```
amessing-bdai, line 18: Should this potentially loop forever or for a fixed number of times then raise an exception? Also, maybe use warn so it will stand out against all the info messages

PR #5, opened 2023-05-03
Reformat the code to black and ruff's standard

ahoelscher-bdai:


jbarry-bdai (approved):
I'm slightly concerned about not doing this on robot because the only unit test is for tf listener but I assume you did your other commit on robot so I'll take it :)

PR #10, opened 2023-06-22
[N/A]  Added spin thread design pattern to NodeWrapper

amessing-bdai:
This PR precedes the spatial math PR.

- Added an option to run a spin thread in the NodeWrapper - this was a design pattern that we were repeating in several places, so hopefully we can use this to reduce the duplicate code. It is set to behave the same as previously by default but has parameters to turn on the new functionality.
- Updated the `TFListenerWrapper` to have a context parameter (Used in unit tests, so we don't get an error from repeated init/shutdown of rclpy)
- Updated the unit test for `TFListenerWrapper` to be multiple tests instead of one lone test function (makes it easier to determine where the test is failing)
- Updated `TFListenerWrapper` to use the new NodeWrapper as it was employing the spin thread pattern
- Updates to the ServiceWrapper were just to make it line up with the changes to NodeWrapper and no changes to its interface
- Also, updated the copyrights to the new format we are using

- [x] I ran HIL tests on spot

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/__init__.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/__init__.py
@@ -1,1 +1,1 @@
-# Copyright [2023] Boston Dynamics AI Institute, Inc.
+# Copyright (c) 2023 Boston Dynamics AI Institute, Inc.  All rights reserved.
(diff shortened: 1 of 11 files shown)


davidwatkins-bdai (approved):
This is a large commit but I couldn't find anything to complain about. Good work.

PR #11, opened 2023-06-23
[N/A] Fixed bug in action client wrapper

amessing-bdai:


ahoelscher-bdai (approved):
Should we get a UT for this in ros_utilities instead of just testing it in our repo?

amessing-bdai:
We should. Currently, there is only testing for the tf_listener_wrapper. I will make a JIRA ticket.

PR #16, opened 2023-08-09
[PN-70] Modifications to ActionClientWrapper for better async handling

amessing-bdai:
1. Exposed NodeWrapper parameters for creating an internal spin thread in ActionClientWrapper
2. Added the option to pass an existing NodeWrapper as an alternative to ActionClientWrapper creating its own
3. Moved ActionHandle (and its associated unit test) and added function to ActionClientWrapper to return ActionHandle
4. Renamed tests to test to match style guide
5. Fixed copyright


mpickett-bdai (approved):
LGTM

PR #17, opened 2023-08-17
[SW-411] Spot teleop is not working as expected

bhung-bdai:
The Spot teleop exposed a slight bug where the robot doesn't time out when it initializes a server that isn't present. This was the error on armless Spots like Fausto. With Gosu, I suspect there's a bigger issue because the Spot Cam payload is causing the driver to fail with this message:
image

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
@@ -21,6 +21,7 @@ class ActionClientWrapper(rclpy.action.ActionClient):
         namespace: Optional[str] = None,
         context: Optional[Context] = None,
         num_executor_threads: int = 1,
+        timeout_sec: float = 10.0,
     ) -> None:
         """Constructor
 
@@ -51,7 +52,7 @@ class ActionClientWrapper(rclpy.action.ActionClient):
         self._node_wrapper.get_logger().info(
             f"Waiting for action server for {self._node.get_namespace()}/{action_name}"
         )
-        self.wait_for_server()
+        self.wait_for_server(timeout_sec)
         self._node_wrapper.get_logger().info("Found server")
 
     def send_goal_and_wait(
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
@@ -52,8 +52,11 @@ class ActionClientWrapper(rclpy.action.ActionClient):
         self._node_wrapper.get_logger().info(
             f"Waiting for action server for {self._node.get_namespace()}/{action_name}"
         )
-        self.wait_for_server(timeout_sec)
-        self._node_wrapper.get_logger().info("Found server")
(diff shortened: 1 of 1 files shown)


ksharma-bdai (commented):
The changes seem reasonable to me, but I'm unclear on what the root cause of the original problem was and if this is a desirable solution. What server is being waited for and under what circumstances do we want this to timeout without error?

bhung-bdai:
You init the robot command and manipulation server in spot_basic.py. It
awaits the manipulation server but doesn't have an arm.

ksharma-bdai:
> You init the robot command and manipulation server in spot_basic.py. It awaits the manipulation server but doesn't have an arm.

Is there any way for `SpotBasic` to handle that more gracefully instead? I'm not arguing either way for this patch, just trying to understand if it's the caller's fault they're waiting for a non-existent service and should be fixed there rather than allowing for semi-transparent passthrough failure here (which could impact other things that you don't want to fail gracefully).

bhung-bdai:
> > You init the robot command and manipulation server in spot_basic.py. It awaits the manipulation server but doesn't have an arm.
> 
> Is there any way for `SpotBasic` to handle that more gracefully instead? I'm not arguing either way for this patch, just trying to understand if it's the caller's fault they're waiting for a non-existent service and should be fixed there rather than allowing for semi-transparent passthrough failure here (which could impact other things that you don't want to fail gracefully).

I think this way works best. Currently there's a limited number of damages that will occur if a server is not found. If you cannot find the corresponding server, then calls to it will respond in a way that will either hang or return nothing currently. Returning nothing doesn't seem to be that bad of a failure case in my opinion. Thoughts, @amessing-bdai?

bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
```diff
@@ -21,6 +21,7 @@ def __init__(
         namespace: Optional[str] = None,
         context: Optional[Context] = None,
         num_executor_threads: int = 1,
+        timeout_sec: float = 10.0,
```
jbarry-bdai, line 24: This should default to the old behavior of waiting indefinitely (-1.0 perhaps?).  There's definitely a workflow where you launch stuff and then let it hang out while you get around to launching something else.

jbarry-bdai (changes requested):
I actually don't think I agree that this should eventually time out.  There's no reason to have a client if the server can't be found.  It also removes a core tenet of ROS2 launch files, which is that it doesn't matter in what order things are launched.

bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
bhung-bdai: I've put a PR fix on the `bdai` repo as an alternative solution, which can be found [here](https://github.com/bdaiinstitute/bdai/pull/700)

bhung-bdai:
I'm going to close this PR with the do not merge label based on what we've discussed. See: [https://github.com/bdaiinstitute/bdai/pull/700](https://github.com/bdaiinstitute/bdai/pull/700)
for the alternative.

PR #18, opened 2023-08-17
[SW-399] Develop unit tests for the service handler 

bhung-bdai:
Ended up also moving the unit tests and service handler into this library to consolidate.


ksharma-bdai (approved):
LGTM, makes sense to move this over here with the other similar classes.

bdai_ros2_wrappers/test/test_service_handle.py
```diff
@@ -0,0 +1,250 @@
...
+    def test_failure(self) -> None:
+        """Tests if the service call is a failure"""
+        self.server_node = FooBarService(self.service_type, self.service_name, callback=_callback_trigger_fail)
+        result, failure = self._internal("trigger_failure")
+        self.assertFalse(result.success)
+        self.assertTrue(failure)
+
+
+class ServiceHandleSetBoolTest(unittest.TestCase):
+    def setUp(self) -> None:
```
baxelrod-bdai, line 137: I see a lot of *similar* `setUp`, `tearDown`, and other functions.  (Not sure how similar or if they are identical).  Is there a way to pull this stuff into something common to avoid the duplication of code?

bdai_ros2_wrappers/test/test_service_handle.py
```diff
@@ -0,0 +1,250 @@
...
+            self._executor.shutdown()
+            self._executor.remove_node(self)
+            self._executor = None
+        if self._thread is not None:
+            self._thread.join()
+            self._thread = None
+        self._service = None
+
+
+class ServiceHandleTriggerTest(unittest.TestCase):
```
baxelrod-bdai, line 78: Its nice to have a human readable description of what exactly each test is checking for.

baxelrod-bdai (approved):

bdai_ros2_wrappers/test/test_service_handle.py
bhung-bdai: I have tried to make them all inherit from a single test case class but I was having some trouble. I can give it a try again and see if there's a specifi way to do it.

baxelrod-bdai: I am more familiar with `TEST_F` in gtest.  Not sure if there is something similar in pytest...

bdai_ros2_wrappers/test/test_service_handle.py
bhung-bdai: Will include

bdai_ros2_wrappers/test/test_service_handle.py
bhung-bdai: Was able to get around it using some sneaky inheritance hacks

PR #19, opened 2023-08-24
[SW-441] Fixing wait for result in the case of a cancel call

abobu-bdai:
Fixed the wait for result return in case of a cancel call.
Still need someone to write unit test.


amessing-bdai:
@bhung-bdai

bhung-bdai (approved):
Now that I've pushed, I think my review is toast. Overall though it seems fine; it might make sense to look at being able to modify `wait_for_result` internally in the future.

amessing-bdai (approved):
Unit test looks good. @bhung-bdai Nice catch about the keyword arg

PR #22, opened 2023-09-11
Enable safe synchronous programming in `rclpy`

mhidalgo-bdai:
Precisely what the title says, based on a conversation with @jbarry-bdai. This PR introduces:

- a multi-threaded `rclpy.executors.Executor` implementation based on a thread pool that scales up and down based on usage, and
- `rclpy.node.Node`, `rclpy_action.client.ActionClient`, and `rclpy_action.server.ActionServer` subclasses to optionally isolate each callback in its own mutually exclusive callback group (effectively precluding concurrent invocations of the same callback)

which, in combination, enable safe synchronous programming across `rclpy` (ie. ROS 2) callbacks without risking a deadlock.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
@@ -49,35 +49,49 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
         if min_workers is None:
             min_workers = 0
         if min_workers < 0:
-            raise ValueError('Mini')
+            raise ValueError(
+                'Minimum number of workers '
+                'must be a nonnegative number')
         self._min_workers = min_workers
 
         if max_workers is None:
             max_workers = 32 * (os.cpu_count() or 1)
         if max_workers < min_workers:
-            raise ValueError('Maximum number of workers must be larger than or equal to the minimum number')
+            raise ValueError(
+                'Maximum number of workers must be larger than'
+                ' or equal to the minimum number of workers')
         self._max_workers = max_workers
 
         if max_idle_time is None:
             max_idle_time = 60.0
         if max_idle_time <= 0:
-            raise ValueError('Maximum idle time for workers must be a positive number')
+            raise ValueError(
+                'Maximum idle time for workers'
+                ' must be a positive number')
         self._max_idle_time = max_idle_time
 
         if submission_quota is None:
             submission_quota = max_workers
         if submission_quota <= 0:
-            raise ValueError('Quotas for submission must be a positive number')
+            raise ValueError(
+                'Quotas for submission '
+                'must be a positive number')
         self._submission_quota = submission_quota
 
         self._submit_lock = threading.Lock()
         self._shutdown_lock = threading.Lock()
 
-        self._workers: typing.List[threading.Thread] = list()
         runqueue: queue.SimpleQueue = queue.SimpleQueue()
         self._runqueue = runqueue
         self._weakref = weakref.ref(
             self, lambda ref: runqueue.put(None))
+        self._workers: typing.List[threading.Thread] = [
+            threading.Thread(
+                target=self._do_work,
+                args=(self._weakref, False),
+                daemon=True
+            ) for _ in range(self._min_workers)
+        ]
         self._waitqueues: typing.Dict[
             typing.Callable[..., typing.Any],
             collections.deque
@@ -104,11 +118,17 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
                 del self._waitqueues[work.fn]
 
     @classmethod
-    def _do_work(cls, executor_weakref: weakref.ref) -> None:
+    def _do_work(
+        cls,
+        executor_weakref: weakref.ref,
+        stop_on_timeout: bool = True
+    ) -> None:
         logger = logging.getLogger(__name__)
         try:
             while True:
-                executor = executor_weakref()
+                executor: typing.Optional[
+                    AutoScalingThreadPool
+                ] = executor_weakref()
                 if executor is None:
                     break
 
@@ -129,6 +149,8 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
                         block=True,
                         timeout=timeout)
                 except queue.Empty:
+                    if stop_on_timeout:
+                        break
                     continue
 
                 if work is not None:
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
@@ -1,3 +1,5 @@
+# Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
+import atexit
 import collections
 import concurrent.futures
 import dataclasses
@@ -12,6 +14,31 @@ import rclpy.executors
 
 
 class AutoScalingThreadPool(concurrent.futures.Executor):
+    """
+    A concurrent.futures.Executor subclass based on a thread pool.
+
+    Akin to the concurrent.futures.ThreadPoolExecutor class but with
+    autoscaling capabilities.
+    """
+
+    _lock: threading.Lock = threading.Lock()
+    _interpreter_shutdown: bool = False
+    _all_runqueues: weakref.WeakSet = weakref.WeakSet()
+    _all_workers: weakref.WeakSet = weakref.WeakSet()
+
+    @classmethod
+    def _on_interpreter_shutdown(cls) -> None:
+        # Gracefully shutdown daemonized threads
+        # upon interpreter shutdown.
+        with cls._lock:
+            cls._interpreter_shutdown = True
+            for queue in cls._all_runqueues:
+                queue.put(None)
+            cls._all_runqueues.clear()
+            for worker in cls._all_workers:
+                if worker.is_alive():
+                    worker.join()
+            cls._all_workers.clear()
 
     @dataclasses.dataclass
     class Work:
@@ -44,8 +71,24 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
         min_workers: typing.Optional[int] = None,
         max_workers: typing.Optional[int] = None,
         submission_quota: typing.Optional[int] = None,
+        submission_patience: typing.Optional[float] = None,
         max_idle_time: typing.Optional[float] = None
     ):
+        """
+        Initializes the thread pool.
+
+        :param min_workers: minimum number of workers in the pool, 0 by default.
+        :param max_workers: maximum number of workers in the pool, 32 times the
+          number of available CPU threads by default (assuming I/O bound work).
+        :param submission_quota: maximum number of concurrent submissions for a
+          a given callable. Up to the maximum number of workers by default. Useful
+          when serving multiple users to prevent anyone from starving the rest.
+        :param submission_patience: time to wait in seconds for a worker to become
+          available before upscaling the pool. 100 ms by default.
+        :param max_idle_time: time in seconds for a worker to wait for work before
+          shutting itself down, effectively downscaling the pool. 60 seconds by default.
+        """
+
         if min_workers is None:
             min_workers = 0
         if min_workers < 0:
@@ -74,43 +117,71 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
             submission_quota = max_workers
         if submission_quota <= 0:
             raise ValueError(
-                'Quotas for submission '
+                'Quota for submission '
                 'must be a positive number')
         self._submission_quota = submission_quota
 
+        if submission_patience is None:
+            submission_patience = 0.1
+        if submission_patience < 0:
+            raise ValueError(
+                'Patience for submission '
+                'must be a nonnegative number')
+        self._submission_patience = submission_patience
+
+        self._shutdown = False
         self._submit_lock = threading.Lock()
         self._shutdown_lock = threading.Lock()
 
-        runqueue: queue.SimpleQueue = queue.SimpleQueue()
-        self._runqueue = runqueue
-        self._weakref = weakref.ref(
-            self, lambda ref: runqueue.put(None))
-        self._workers: typing.List[threading.Thread] = [
-            threading.Thread(
-                target=self._do_work,
-                args=(self._weakref, False),
-                daemon=True
-            ) for _ in range(self._min_workers)
-        ]
+        with AutoScalingThreadPool._lock:
+            if AutoScalingThreadPool._interpreter_shutdown:
+                raise RuntimeError(
+                    'cannot create thread pool while'
+                    ' interpreter is shutting down')
+            # register runqueue for external wake up  on interpreter shutdown
+            runqueue: queue.SimpleQueue = queue.SimpleQueue()
+            AutoScalingThreadPool._all_runqueues.add(runqueue)
+            self._runqueue = runqueue
+            weak_self = weakref.ref(
+                self, lambda ref: runqueue.put(None))
+            self._workers: weakref.WeakSet = weakref.WeakSet()
+            for _ in range(self._min_workers):  # fire up stable worker pool
+                # careful with back references to self!
+                worker = threading.Thread(
+                    target=AutoScalingThreadPool._do_work,
+                    args=(weak_self, False), daemon=True)
+                worker.start()
+                self._workers.add(worker)
+            # register workers for external joining on interpreter shutdown
+            AutoScalingThreadPool._all_workers.update(self._workers)
+
         self._waitqueues: typing.Dict[
             typing.Callable[..., typing.Any],
             collections.deque
         ] = collections.defaultdict(collections.deque)
         self._runlists: typing.Dict[
-            typing.Callable[..., typing.Any], set
-        ] = collections.defaultdict(set)
+            typing.Callable[..., typing.Any], list
+        ] = collections.defaultdict(list)
         self._runslots = threading.Semaphore(0)
+
+    @property
+    def workers(self) -> typing.List[threading.Thread]:
+        with self._submit_lock:
+            # avoid races and proactively drop shutdown workers
+            self._workers = weakref.WeakSet([
+                w for w in self._workers if w.is_alive()])
+        return list(self._workers)
 
     def _do_cleanup_after(self, work: 'AutoScalingThreadPool.Work') -> None:
         with self._submit_lock:
             self._runlists[work.fn].remove(work)
-            if self._waitqueues[work.fn]:
+            if self._waitqueues[work.fn]:  # continue with pending work
                 work = self._waitqueues[work.fn].popleft()
                 while work.cancelled() and self._waitqueues[work.fn]:
                     work = self._waitqueues[work.fn].popleft()
                 if not work.cancelled():
                     self._do_submit(work)  # actually submit work
-                    self._runlists[work.fn].add(work)
+                    self._runlists[work.fn].append(work)
                     work.when_done(self._do_cleanup_after)
             if not self._runlists[work.fn]:
                 del self._runlists[work.fn]
@@ -125,7 +196,7 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
     ) -> None:
         logger = logging.getLogger(__name__)
         try:
-            while True:
+            while not cls._interpreter_shutdown:
                 executor: typing.Optional[
                     AutoScalingThreadPool
                 ] = executor_weakref()
@@ -141,7 +212,7 @@ class AutoScalingThreadPool(concurrent.futures.Executor):
                     runqueue.put(None)
                     break
 
-                del executor
+                del executor  # drop reference
 
                 runslots.release()
                 try:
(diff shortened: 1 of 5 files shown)


mhidalgo-bdai:
Uh, `pre-commit` didn't complain on local `git commit`. Style issues fixed in https://github.com/bdaiinstitute/ros_utilities/pull/22/commits/b96a16a3295026fec3f4d11f7b55af543ed4801e.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,409 @@
...
+        if min_workers is None:
+            min_workers = 0
+        if min_workers < 0:
+            raise ValueError(
+                'Minimum number of workers '
+                'must be a nonnegative number')
+        self._min_workers = min_workers
+
+        if max_workers is None:
+            max_workers = 32 * (os.cpu_count() or 1)
```
mhidalgo-bdai, line 101: @jbarry-bdai on a typical Intel CPU with 4 physical cores, 8 logical threads, this amounts to 512 concurrent threads. That should be enough for most use cases. We can lift the restriction entirely too, but that many concurrent threads would suggest other issues.

mhidalgo-bdai:
Silly me, `pre-commit` hooks were not installed. Now, done complying in https://github.com/bdaiinstitute/ros_utilities/pull/22/commits/76c066568f1e59f9fd46a5e2ce4d85a9c22c9d99.

mhidalgo-bdai:
Ended up sticking to `pytest` rather than `launch_pytest` as it the test reads better. I also pushed a few fixes as tests uncovered some concurrency related booby traps.

bdai_ros2_wrappers/test/test_integration.py
```diff
@@ -0,0 +1,162 @@
...
+    ros_executor.add_node(ros_node)
+    try:
+        yield
+    finally:
+        for node in ros_executor.get_nodes():
+            ros_executor.remove_node(node)
+
+
+def test_blocking_sequence(ros_executor: Executor, ros_node: Node) -> None:
+    """
```
mhidalgo-bdai, line 121: FYI @jbarry-bdai @amessing-bdai is something like this that you had in mind when discussing a "simpler" programming model?

mhidalgo-bdai:
I think that, with this model, we can reduce the LOC count of this package substantially as fewer wrappers are needed.

amessing-bdai:
Overall this looks great! Can you show an example of running an `ActionClient.send_goal` inside the callback of an `ActionServer` or `Subscriber`?

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,395 @@
...
+                    complete = False
+                else:
+                    work.notify_cancelation()
+            if not self._runlists[work.fn]:
+                del self._runlists[work.fn]
+        return complete
+
+    @classmethod
+    def _do_work(cls, executor_weakref: weakref.ref, stop_on_timeout: bool = True) -> None:
+        logger = logging.getLogger(__name__)
```
amessing-bdai, line 198: How does this interact with RcUtilLogger and more specifically does it get captured by rosbags?

mhidalgo-bdai: It doesn't interact at all right now. We can make it so, however, by forwarding a logger instance on pool instantiation.

mhidalgo-bdai: Done in https://github.com/bdaiinstitute/ros_utilities/pull/22/commits/2ee1d41e5649c17cd8e75d8b4df02c4f70c99906.

mhidalgo-bdai:
> Can you show an example of running an ActionClient.send_goal inside the callback of an ActionServer or Subscriber?

@amessing-bdai I chained a service call with two action goals in https://github.com/bdaiinstitute/ros_utilities/pull/22/commits/7478b0bbf14c34982a063df7159144a56c402ca8.

mhidalgo-bdai:
It occurred to me last night that a (custom) non-reentrant callback group would be lighter alternative to having a mutually exclusive callback group per callback. It's also more likely to be accepted upstream. See https://github.com/bdaiinstitute/ros_utilities/pull/22/commits/53ef806f67d25144423aaad8d9b8b0077d664c47. As a bonus, it simplifies code.

mhidalgo-bdai:
FYI there are problems with the pre-existing `test_action_handle.py`. It sometimes hangs on CI, and it's generating future related warnings locally. We could try to fix it, but I'd be inclined towards simplifying all the threading that's going on there with the new functionality this PR brings.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
jbarry-bdai: Haha yes this seems good.  The point is to make it easy for the general/casual user.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+            self.future.set_running_or_notify_cancel()
+
+    def __init__(
+        self,
+        *,
+        min_workers: typing.Optional[int] = None,
+        max_workers: typing.Optional[int] = None,
+        submission_quota: typing.Optional[int] = None,
+        submission_patience: typing.Optional[float] = None,
+        max_idle_time: typing.Optional[float] = None,
```
jbarry-bdai, line 80: Why have the integral types be optional instead of set to their default values?

bdai_ros2_wrappers/bdai_ros2_wrappers/node.py
```diff
@@ -87,3 +89,21 @@ def shutdown(self) -> None:
     # def __del__(self) -> None:
     #     """Destructor"""
     #     self.shutdown()
+
+
+class FriendlyNode(Node):
```
jbarry-bdai, line 94: Maybe "NonReentrantNode" or "AutoScalingNode"?  Friendly is a pretty awesome name but doesn't exactly make it obvious it goes with the other stuff :)

bdai_ros2_wrappers/test/test_integration.py
jbarry-bdai: Yes I think so.

jbarry-bdai (commented):
I like where this is going.  My thought is we should go ahead and merge the addition and then immediately start removing the existing wrappers so as not to have too long where both options exist but also not enormous PRs.  I could be talked into doing this addition and the deletion of the code it makes redundant in one PR though.

jbarry-bdai:
I think we do need additions to the `README` (if it exists - otherwise creating one) and docstrings before this is ready to go but otherwise LGTM.

bdai_ros2_wrappers/bdai_ros2_wrappers/node.py
mhidalgo-bdai: Yeah, `FriendlyName` is a throwaway name :sweat_smile:

We could just call it `Node`. That way downstream code can switch from one to the other by changing the import statements ie. from `from rclpy.node import Node` to `from bdai_ros.node import Node`.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
mhidalgo-bdai: A bit of habit mixed with liking the idea of making it explicitly optional through type annotations. So no strong reasons.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+                    work = self._waitqueues[work.fn].popleft()
+                if not self._waitqueues[work.fn]:
+                    del self._waitqueues[work.fn]
+                if not work.cancelled():
+                    self._runlists[work.fn].append(work)
+                    self._runqueue.put(work)  # actually submit work
+                    complete = False
+                else:
+                    work.notify_cancelation()
+            if not self._runlists[work.fn]:
```
bhung-bdai, line 204: Same for this one. I'm not sure if it's a hard guarantee or not, so feel free to correct me.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+                    work.execute()
+        except BaseException:
+            import textwrap
+            import traceback
+
+            logger.error("Worker threw an exception: \n%s", textwrap.indent(traceback.format_exc(), "  "))
+
+    def _do_submit(self, work: "AutoScalingThreadPool.Work") -> None:
+        self._workers = weakref.WeakSet([w for w in self._workers if w.is_alive()])
+        while not self._runslots.acquire(timeout=self._submission_patience):
```
bhung-bdai, line 251: I'm a bit confused about this part. If the semaphore is full, this loop spins up another thread to run the new function right? Does the counter on the semaphore or number of things that can access the semaphore get updated accordingly?

bhung-bdai (commented):
Some questions for my own personal understanding. This looks very neat and should help a lot!

bdai_ros2_wrappers/bdai_ros2_wrappers/callback_groups.py
```diff
@@ -0,0 +1,38 @@
...
+import rclpy.callback_groups
+import rclpy.executors
+
+
+class NonReentrantCallbackGroup(rclpy.callback_groups.CallbackGroup):
+    """
+    A callback group to prevent concurrent execution of the same callback.
+
+    Concurrent execution of different callbacks is allowed.
+    """
```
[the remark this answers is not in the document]
mhidalgo-bdai: That is correct. I extended the documentation in 3140d4435086f57dae54c5d5a17b7c7a6ee8fd7a to make the distinction clearer.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
mhidalgo-bdai: > what does tuning actually look like for this for say an embedded low powered platform? Is there a way to stress test this end to end through ROS?

That is a great question I don't have a good answer for. One would think that the size of the pool would normally depend on latency requirements considering incoming data rates and processing rates, constrained by the resources you have. Blocking calls across callbacks adds the maximum depth of the associated call graphs to the set of affecting factors. So I'd probably start with these defaults, and guess my way out of trouble if it ever comes to that. You _could_ do a more thorough analysis by heavily tracing your system, and [ROS 2 has tracing support](https://github.com/ros2/ros2_tracing) for it. Not trivial though.

And then you have the _other_ parameters in this executor. Quotas are really just a way to mitigate a sudden imbalance between data rates and processing rates for reentrant callback groups (ie. for some reason you don't process things fast enough and the number of callbacks starts to pile up). You may not have any, which is why they are disabled by default (by making them equal to the maximum pool size). Patience is a bound on OS overhead for firing up threads, so OS dependent.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+        if max_idle_time is None:
+            max_idle_time = 60.0
+        if max_idle_time <= 0:
+            raise ValueError("Maximum idle time for workers must be a positive number")
+        self._max_idle_time = max_idle_time
+
+        if submission_quota is None:
+            submission_quota = max_workers
+        if submission_quota <= 0:
+            raise ValueError("Quota for submission must be a positive number")
```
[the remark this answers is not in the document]
mhidalgo-bdai: Hmm, [my understanding](https://en.wikipedia.org/wiki/Sign_(mathematics)#:~:text=Terminology%20for%20signs,-When%200%20is&text=A%20number%20is%20positive%20if,than%20or%20equal%20to%20zero.) is that positive means `x > 0` and nonnegative means `x > 0`, in which case these are correct?

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+
+import rclpy.executors
+
+
+def classname(obj: typing.Any) -> str:
+    cls = obj.__class__
+    return f"{cls.__module__}.{cls.__qualname__}"
+
+
+class AutoScalingThreadPool(concurrent.futures.Executor):
```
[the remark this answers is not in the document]
mhidalgo-bdai: > I think all of this logic makes sense, but I am concerned about future debug-ability and getting stuck unexpectedly in the future

> I may just be overly paranoid, but it took me a little bit on paper to walk through things and I probably need to do another pass.

And you're a wise man for it. While adding more tests I uncovered some race conditions :skull_and_crossbones: . I had to change a few things to fix them. PTAL.

> It would be nice to have debug statements that can be enabled/disabled 

I added debug logs all over the place in f014606917ba29ce24713a1f617a0fa9d72bfb9f.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+        future: concurrent.futures.Future
+        fn: typing.Callable[..., typing.Any]
+        args: typing.Tuple[typing.Any, ...]
+        kwargs: typing.Dict[str, typing.Any]
+
+        def execute(self) -> None:
+            if not self.future.set_running_or_notify_cancel():
+                return
+            try:
+                self.future.set_result(self.fn(*self.args, **self.kwargs))
```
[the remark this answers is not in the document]
mhidalgo-bdai: Right, it would be an issue, but it cannot happen in this implementation. Cancellations are notified for work that is cancelled while in _waitqueues_, while work is still out of reach for workers that can execute it. I've extended the documentation to discuss these waitqueues.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+        self._shutdown_lock = threading.Lock()
+
+        with AutoScalingThreadPool._lock:
+            if AutoScalingThreadPool._interpreter_shutdown:
+                raise RuntimeError("cannot create thread pool while interpreter is shutting down")
+            # register runqueue for external wake up  on interpreter shutdown
+            runqueue: queue.SimpleQueue = queue.SimpleQueue()
+            AutoScalingThreadPool._all_runqueues.add(runqueue)
+            self._runqueue = runqueue
+            weak_self = weakref.ref(self, lambda ref: runqueue.put(None))
```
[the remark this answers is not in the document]
mhidalgo-bdai: > is there any chance that any of these weak references get eaten by the garbage collection before the job is done running

It could happen, yes, but that's a non-issue. Once the worker is done working, if the executor is gone, it will self-terminate.

> What happens to the weak_self in the thread after the _do_work function runs?

Things changed a bit since you reviewed, but not fundamentally. We keep a weak reference to the executor to avoid the reference from persisting the pool in a daemonized worker once it can no longer be accessed from anywhere else. We dereference it when necessary and delete the reference explicitly for the same reasons.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,414 @@
...
+    def _cleanup_after(self, work: "AutoScalingThreadPool.Work") -> bool:
+        complete = True
+        with self._submit_lock:
+            self._runlists[work.fn].remove(work)
+            if work.fn in self._waitqueues and self._waitqueues[work.fn]:  # continue with pending work
+                work = self._waitqueues[work.fn].popleft()
+                while work.cancelled() and self._waitqueues[work.fn]:
+                    work.notify_cancelation()
+                    work = self._waitqueues[work.fn].popleft()
+                if not self._waitqueues[work.fn]:
```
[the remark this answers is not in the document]
mhidalgo-bdai: Not `None` but an empty `deque`.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
mhidalgo-bdai: In this case it is an empty `list`.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
mhidalgo-bdai: Yes, the semaphore ie. the runslot count gets increased by workers. Submitting work 
decreases the count, workers ready to work increase it. I added documentation to clarify this.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,610 @@
...
+                break
+            self._logger.debug("Upscaling pool...")
+            with self._scaling_event:
+                worker = AutoScalingThreadPool.Worker(self._weak_self)
+                with AutoScalingThreadPool._lock:
+                    AutoScalingThreadPool._all_workers.add(worker)
+                self._workers.add(worker)
+                self._scaling_event.notify_all()
+            self._logger.debug("Done upscaling")
+        self._logger.debug("Got worker!")
```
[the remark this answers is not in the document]
mhidalgo-bdai: Yeah, strictly speaking you don't know. I'll remove it.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,610 @@
...
+                    del executor  # drop reference
+
+                    self._logger.debug("Worker waiting for work...")
+                    try:
+                        work = self._runqueue.get(block=True, timeout=self._timeout)
+                    except queue.Empty:
+                        self._logger.debug("Worker timed out waiting!")
+                        if runslots.acquire(blocking=False):
+                            self._logger.debug("Terminating worker...")
+                            break
```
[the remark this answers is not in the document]
mhidalgo-bdai: > this is the termination mechansim for workers when a timeout is specified?

Correct.

> no timeout workers will not get here and will only break out on shutdown when None is added to the queue?

Correct.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,610 @@
...
+            self._logger.debug(f"Cleaning up after work '{work}'")
+            self._runlists[work.fn].remove(work)
+            if work.fn in self._waitqueues and self._waitqueues[work.fn]:  # continue with pending work
+                self._logger.debug("Have similar work pending!")
+                self._logger.debug("Fetching pending work...")
+                work = self._waitqueues[work.fn].popleft()
+                while work.cancelled() and self._waitqueues[work.fn]:
+                    self._logger.debug(f"Work '{work}' was cancelled, notify it")
+                    work.notify_cancelation()
+                    work = self._waitqueues[work.fn].popleft()
```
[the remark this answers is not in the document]
mhidalgo-bdai: Access to the waitqueues is synchronized by the `submit_lock`, so if we get to this point then we know the waitqueue is non-empty. There is check below, outside the loop, to drop cancelled work. So we burn through the waitqueue looking for non-cancelled work, stick to the last piece of work regardless, and drop it if turns out to have been cancelled.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,610 @@
...
+                if work.fn in self._waitqueues and self._waitqueues[work.fn]:  # prioritize pending work
+                    self._logger.debug("Have similar work pending")
+                    self._logger.debug(f"Work '{work}' put to wait", work)
+                    self._waitqueues[work.fn].append(work)
+                    self._logger.debug("Fetching pending work...")
+                    work = self._waitqueues[work.fn].popleft()
+                    while work.cancelled() and self._waitqueues[work.fn]:
+                        self._logger.debug(f"Work '{work}' was cancelled, notify it")
+                        work.notify_cancelation()
+                        work = self._waitqueues[work.fn].popleft()
```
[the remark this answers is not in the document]
mhidalgo-bdai: Hmm, yeah, it could be split into its own method. I didn't for the sake of keeping code as clear as possible, but we can.

bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
```diff
@@ -0,0 +1,610 @@
...
+        with self._shutdown_lock:
+            self._shutdown = True
+        self._logger.debug("Pool shutdown")
+
+        if cancel_futures:
+            with self._submit_lock:
+                self._logger.debug("Canceling all work in progress...")
+                for runlist in self._runlists.values():
+                    for work in runlist:
+                        work.cancel()
```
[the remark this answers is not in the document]
mhidalgo-bdai: Fortunately, [it is an idempotent method](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel). Checking wouldn't work anyways, as it would be subject to TOCTOU races.

PR #24, opened 2023-09-18
[PN-64] Add ROSBag_Recorder Utility

abaker-bdai:
Adds a wrapper around `ros2bag::record` to simplify recording rosbags when creating python nodes

This will be followed up by a PR to `bdai` that will update/add examples on how to use the utility with the walk forward example

In order to test, refer to the [`example_walk_forward_with_recording.py`](https://github.com/bdaiinstitute/bdai/blob/ffaa88b5f2c5edb611f441a002b7f794da7f779d/ws/src/spot_utilities/examples/example_walk_forward_with_recording.py) from [this PR to bdai](https://github.com/bdaiinstitute/bdai/pull/921)

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/rosbag_recorder.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/rosbag_recorder.py
@@ -1,16 +1,19 @@
 # Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
 import threading
+from pathlib import Path
 
 import rclpy
+from rclpy.impl import rcutils_logger
 from rosbag2_py import Recorder, RecordOptions, StorageOptions
 
 
 class ROSBagRecorder(object):
     def __init__(
         self,
-        bag_file_path: str,
+        bag_file_path: Path,
         storage_id: str = "sqlite3",
     ) -> None:
+        self._logger = rcutils_logger.RcutilsLogger(name="rosbag_recorder")
         self._rosbag_recorder = Recorder()
 
         self._record_options = RecordOptions()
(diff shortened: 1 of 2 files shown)


bdai_ros2_wrappers/test/test_rosbag_recorder.py
```diff
@@ -0,0 +1,63 @@
...
+import os
+import unittest
+
+import rclpy
+from rclpy.node import Node
+from std_msgs.msg import String
+
+from bdai_ros2_wrappers.rosbag_recorder import ROSBagRecorder
+
+BAG_FILE_DIR = os.path.join(os.getenv("BDAI", "/workspaces/bdai"), "recording")
```
amessing-bdai, line 11: This is an open-source repo for those outside of bdai, so can you set up the test assuming it is not being run inside the docker

abaker-bdai: Done!

bdai_ros2_wrappers/bdai_ros2_wrappers/rosbag_recorder.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def stop(self) -> None:
+        self._logger.debug("Stopping Rosbag recorder")
+        self._rosbag_recorder.cancel()
+        self._thread.join()
+
+    def _record(self) -> None:
+        if self._bag_path.exists():
+            self._logger.error("Cannot overwrite rosbag that already exists.")
+            return
```
tperkins-bdai, line 44: Looks good! My only question is, if the bag already exists, how does this fail? This looks like it just exits the thread, but it would be ideal if it caused `__init__` or `start` to raise an exception.

abaker-bdai:
Closing in favor of [this PR to `bdai`](https://github.com/bdaiinstitute/bdai/pull/921) where `ROSBagRecorder` will now be implemented as a singleton in `record.py`

PR #25, opened 2023-09-22
Push ROS 2 to the background

mhidalgo-bdai:
Owing to the nature of most indirect and direct use of these wrappers in BDAII's codebase, this PR re-introduces the concept of a process-wide node, served in the background.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/process.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/process.py
@@ -1,0 +1,217 @@
+# Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
+import functools
+import inspect
+import pathlib
+import sys
+import threading
+import typing
+
+import rclpy
+import rclpy.executors
+import rclpy.node
+
+from bdai_ros2_wrappers.executors import AutoScalingMultiThreadedExecutor
+from bdai_ros2_wrappers.node import Node
+
+
+class ROSAwareScope(typing.ContextManager["ROSAwareScope"]):
+    """
+    A context manager to enable ROS code in a given scope by providing
+    an `rclpy` node and an autoscaling multi-threaded executor spinning
+    in a background thread.
+    """
+
+    def __init__(
+        self,
+        name: str,
+        context: typing.Optional[rclpy.context.Context] = None,
+        node_args: typing.Optional[typing.Dict[str, typing.Any]] = None,
+        executor_args: typing.Optional[typing.Dict[str, typing.Any]] = None,
+    ):
+        """
+        Initializes the ROS aware scope.
+
+        Args:
+            name: name for this scope, and for the underlying node.
+            context: optional context for the underlying node.
+            node_args: optional keyword arguments for the underlying node.
+            executor_args: optional keyword arguments for the underlying executor.
+        """
+        self._name = name
+        self._context = context
+        self._node: typing.Optional[rclpy.node.Node] = None
+        self._node_args: typing.Dict[str, typing.Any] = node_args or {}
+        self._executor: typing.Optional[rclpy.executors.Executor] = None
+        self._executor_args: typing.Dict[str, typing.Any] = executor_args or {}
+        self._background_thread: typing.Optional[threading.Thread] = None
+        self._active = False
+
+    def __enter__(self) -> "ROSAwareScope":
+        """
+        Activates scope.
+
+        An active scope cannot be re-entered.
+        """
+        if self._active:
+            raise RuntimeError("cannot enter active scope")
+        self._node = Node(self._name, context=self._context, **self._node_args)
+        try:
+            self._executor = AutoScalingMultiThreadedExecutor(
+                context=self._context, logger=self._node.get_logger(), **self._executor_args
+            )
+            self._background_thread = threading.Thread(target=self._executor.spin)
+            try:
+                self._executor.add_node(self._node)
+                self._background_thread.start()
+            except:
+                self._executor.shutdown()
+                self._executor = None
+                self._background_thread = None
+                raise
+            self._active = True
+            return self
+        except:
+            self._node.destroy_node()
+            self._node = None
+            raise
+
+    def __exit__(self, *exc: typing.Any) -> None:
+        """
+        Deactivates scope.
+
+        An inactive scope cannot be deactivated.
+        """
+        if not self._active:
+            raise RuntimeError("cannot exit inactive scope")
+        assert self._executor is not None
+        assert self._background_thread is not None
+        assert self._node is not None
+        self._executor.shutdown()
+        self._executor = None
+        self._background_thread.join()
+        self._background_thread = None
+        self._node.destroy_node()
+        self._node = None
+        self._active = False
+
+    @property
+    def node(self) -> rclpy.node.Node:
+        """Scope node, if active."""
+        if not self._active:
+            raise RuntimeError("scope is not active")
+        assert self._node is not None
+        return self._node
+
+    @property
+    def executor(self) -> rclpy.executors.Executor:
+        """Scope executor, if active."""
+        if not self._active:
+            raise RuntimeError("scope is not active")
+        assert self._executor is not None
+        return self._executor
+
+
+MainCallableWithArgs = typing.Callable[[typing.Optional[typing.Sequence[str]]], typing.Optional[int]]
+MainCallableWithNoArgs = typing.Callable[[], typing.Optional[int]]
+MainCallable = typing.Union[MainCallableWithNoArgs, MainCallableWithArgs]
+
+
+class ROSAwareProcess:
+    """
+    A ``main``-like function wrapper that binds a ROS aware scope to each function invocation.
+
+    Only one process can be invoked at a time, becoming the ``current`` process.
+    This enables global access to the process and, in consequence, to its scope.
+
+    This wrapper will rarely be used explicitly but through the `wrap` decorator.
+    """
+
+    lock = threading.Lock()
+
+    current: typing.Optional["ROSAwareProcess"] = None
+
+    def __init__(self, func: MainCallable, name: typing.Optional[str] = None, **kwargs: typing.Any):
+        """
+        Initializes the ROS aware process.
+
+        Args:
+            func: a ``main``-like function to wrap.
+            name: an optional name for this process and the underlying node.
+            If not provided, the basename without extension of the calling executable will be used.
+
+        Additional keyword arguments will be forwarded to the process' `ROSAwareScope` instance.
+        """
+        self._func = func
+        if name is None:
+            name = pathlib.Path(sys.argv[0]).stem
+        self._name = name
+        self._kwargs = kwargs
+        self._scope: typing.Optional[ROSAwareScope] = None
+        functools.update_wrapper(self, self._func)
+
+    def __getattr__(self, name: str) -> typing.Any:
+        """Gets attribute from scope, if any."""
+        if self._scope is None:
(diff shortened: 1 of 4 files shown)


mhidalgo-bdai:
@jbarry-bdai @amessing-bdai for the moment, this only demonstrates how `TFListenerWrapper` would change. Once you're OK with the API, I'll migrate the rest to rely on this new logic.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -4,25 +4,24 @@
...
-        self,
-        node_name: str,
-        wait_for_transform: Optional[Tuple[str, str]] = None,
-        cache_time_s: Optional[int] = None,
-        context: Optional[Context] = None,
+        self, node: Node, wait_for_transform: Optional[Tuple[str, str]] = None, cache_time_s: Optional[int] = None
     ) -> None:
-        # private because we want to make sure no one but us spins this node!
-        self._node = NodeWrapper(node_name=node_name, context=context, spin_thread=True)
+        node = node or process.node()
```
bhung-bdai, line 20: Evaluation to a Boolean?

bdai_ros2_wrappers/bdai_ros2_wrappers/process.py
```diff
@@ -0,0 +1,217 @@
...
+                with ROSAwareScope(self._name, **self._kwargs) as scope:
+                    self._scope = scope
+                    ROSAwareProcess.current = self
+                    try:
+                        sig = inspect.signature(self._func)
+                        if not sig.parameters:
+                            func_no_args = typing.cast(MainCallableWithNoArgs, self._func)
+                            return func_no_args()
+                        func_with = typing.cast(MainCallableWithArgs, self._func)
+                        return func_with(rclpy.utilities.remove_ros_args(args))
```
bhung-bdai, line 185: If this function throws an exception or error, is there a way to easily recover the thread or will it just exit? Asking because I see a case where something can die ungracefully in this thread with something else outside the scope depending on it (if I am reading this correctly).

bhung-bdai (commented):
Some general questions. I am admittedly confused on what exactly the score of the ROSAware thread is supposed to cover and when the appropriate time to use it is.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -4,25 +4,24 @@
...
 
 
 class TFListenerWrapper(object):
     def __init__(
-        self,
-        node_name: str,
-        wait_for_transform: Optional[Tuple[str, str]] = None,
-        cache_time_s: Optional[int] = None,
-        context: Optional[Context] = None,
+        self, node: Node, wait_for_transform: Optional[Tuple[str, str]] = None, cache_time_s: Optional[int] = None
```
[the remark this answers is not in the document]
mhidalgo-bdai: > is the node argument here primarily for the test cases or is there a use case where that would be appropriate? it seems like a slight departure from the original code there

For test cases and for multi-node use cases. It does depart from the original code because it is no longer a node of its own.

> should that be typed/default as Optional[Node] = None to make it more clear?

Yes! Fixed in https://github.com/bdaiinstitute/ros_utilities/pull/25/commits/82f8445e40905303f70ec7cb908994d833fbb0ea.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
mhidalgo-bdai: This one's a handy gimmick. Values are coerced to boolean by the operator, but the expression yields the original value. So `None or 'test'` yields `'test'`.

bdai_ros2_wrappers/bdai_ros2_wrappers/process.py
mhidalgo-bdai: First of all, bear in mind that this is designed to decorate top level functions e.g. `main`. If an exception manages to propagate all the way up there, there's nothing beyond but the interpreter. 

It could happen that the thread that "owns" the node dies while another thread that "borrows" it remains alive, but I think all of those can be labelled as API misuse.

bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
```diff
@@ -1,58 +1,33 @@
...
-            )
-        else:
-            raise TypeError("`node` should be either a `NodeWrapper` or `str`")
-        super().__init__(self._node_wrapper, action_type, action_name)
-        self._node_wrapper.get_logger().info(
-            f"Waiting for action server for {self._node.get_namespace()}/{action_name}"
-        )
+        node = node or process.node()
+        if node is None:
+            raise ValueError("no ROS 2 node available")
```
jbarry-bdai, line 25: Should this perhaps tell the user a little more - like "To use the default process node make sure you are using X executor" or something?

bdai_ros2_wrappers/bdai_ros2_wrappers/action_handle.py
```diff
@@ -128,29 +140,21 @@ def get_feedback_callback(self, feedback: Action.Feedback) -> None:
...
-            if self._on_cancel_success_callback is not None:
-                self._on_cancel_success_callback()
-            return
+        self._exception = future.exception()
+        if self._exception is None:
+            self._result = future.result()
+            if self._result.status == GoalStatus.STATUS_SUCCEEDED:
+                self._logger.info("Finished successfully")
+                if self._result_callback is not None:
+                    self._result_callback(self._result.result)
```
jbarry-bdai, line 149: This confuses me and did in the earlier instantiation of this as well.  If it's called `result_callback` shouldn't it be run on the return of a result whether or not it's successful?  Should the name actually be `success_callback`?  @amessing-bdai

bdai_ros2_wrappers/bdai_ros2_wrappers/service_client.py
```diff
@@ -1,63 +0,0 @@
-# Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
```
jbarry-bdai, line 1: Was this removed because now just doing `service.call` with a native client won't deadlock?

mhidalgo-bdai:
> Are there any cases where you would see us not wanting to use this a default behavior in other areas of the code? e.g. is there a case where an explicitly managed node with a single threaded executor would make sense or should the auto-scaling executor behavior be essentially identical and equally usable in all cases?

@ksharma-bdai Good question. I think it goes back to the sane defaults discussion. One could start out with these "process-wide" nodes, and defer complex setups until necessary. Like for any form of software, large or complicated units of functionality are better split into smaller, single-purpose, and hopefully reusable ones. Having closely related nodes in a single process enables performance optimizations (even if they are not available just yet). If determinism and/or scalability is a requirement, single-threaded executors are better suited than multi-threaded ones (even if stock ones are not useful for it just yet).

I know there is a lot of future-proofing (and wishful thinking?) going around. It still holds conceptually.

bdai_ros2_wrappers/bdai_ros2_wrappers/service_client.py
mhidalgo-bdai: Correct.

bdai_ros2_wrappers/bdai_ros2_wrappers/action_handle.py
amessing-bdai: Probably should to be more clear. While we are discussing the specifics, it might also make sense to have different callbacks for rejected and aborted. Information can be passed back in `future.result().result` when an action is aborted. This information could be useful to a user.

mhidalgo-bdai:
Alright, there we go. Fixed all races in https://github.com/bdaiinstitute/ros_utilities/pull/25/commits/50697e47ba8e5eed624e56c0a3873a5a0f85ebd0.

bdai_ros2_wrappers/bdai_ros2_wrappers/action_handle.py
mhidalgo-bdai: FTR I tried to keep APIs unchanged as much as possible to avoid major surgery downstream. In some cases I had to make changes as the implementation was inherently race-y. If we are up for an overhaul, I have some ideas :upside_down_face:

bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
mhidalgo-bdai: Note however that this idea of a default process node is orthogonal to that of the executor. We could be using a single-threaded executor in the background (with all the limitations you already know about).

bdai_ros2_wrappers/bdai_ros2_wrappers/action_handle.py
amessing-bdai: AFAIK, SpotBasic is the only thing that uses this, so I am good with an overhaul. Really, we just want an easy way to deal with the lifecycle of an Action from the client side, an easy way to set specific callbacks, and overall to reduce duplication of boilerplate code from working with actions.

mhidalgo-bdai: Right, but that new API is likely to percolate up to `SpotBasic`, changing SpotBasic API itself. That's why I held it back.

mhidalgo-bdai:
Added https://github.com/bdaiinstitute/ros_utilities/pull/25/commits/8c1ee74f6bbddb403cb7d224eac840d8cbf2083c to ease migration downstream.

bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
amessing-bdai: That's fair. I agree with Jenny that we should have a suggestion for how to fix the error. Something like "please provide a node or use the process-wide node via _".

bdai_ros2_wrappers/bdai_ros2_wrappers/action_handle.py
amessing-bdai: Do you mean the internals or the actual API? If it is just the internals and there is no regression, then I am still good with proceeding. If you are changing the actual API then we will need to discuss it. My current understanding is the constructor would change, but the actual function declarations would remain the same (based on https://github.com/bdaiinstitute/bdai/pull/927)?

mhidalgo-bdai: > My current understanding is the constructor would change, but the actual function declarations would remain the same (based on https://github.com/bdaiinstitute/bdai/pull/927)?

That is, if we keep APIs they way they are now. By API overhaul I mean, for example, changing how we deal with ongoing actions to be less callback-y and embrace asynchronous/synchronous programming. I left some ideas on the internal ticket behind this PR. We can continue the discussion there. No need to go forth with big changes here and now.

bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
mhidalgo-bdai: See https://github.com/bdaiinstitute/ros_utilities/pull/25/commits/30ad62da27fefa15637b9dc2063a5aa4b5c2f3f6.

mhidalgo-bdai:
Alright, this is big enough already. I'd rather review and merge this one, then review and merge the rest. @amessing-bdai @jbarry-bdai @ksharma-bdai PTAL.

amessing-bdai (approved):
LG and I think all questions have been answered

PR #28, opened 2023-10-02
Log forwarding for process

mhidalgo-bdai:
This patch leverages #23 to add automatic log forwarding whenever using process-wide ROS 2 machinery. Builds on #25.


bhung-bdai (approved):
LGTM. Does this break any of the previous tests you had set up for `process.py`?

mhidalgo-bdai:
> Does this break any of the previous tests you had set up for process.py?

It does not, this feature is orthogonal to the rest. Errors would've shown during CI too.

mhidalgo-bdai:
Going in to continue building on top.

PR #29, opened 2023-10-02
Add `wait_for_shutdown` utility

mhidalgo-bdai:
This patch brings a `wait_for_shutdown` utility to wait for a given context to shutdown. Useful to block the main thread while callbacks are invoked in the background. Builds on #25.


bhung-bdai (commented):
Similar question to the other threaded PRs: does this shut down only the work being done on that thread, or does it shut down the executor globally?

mhidalgo-bdai:
> does this shut down only the work being done on that thread, or does it shut down the executor globally?

`wait_for_shutdown` does not shut down anything, it only performs a blocking wait until shutdown or timeout. As it's blocking, shutdown can only happen from another thread or signal handler. It serves the same purpose as [`rospy.spin`](http://wiki.ros.org/rospy/Overview/Initialization%20and%20Shutdown) did: just hold until ROS 2 shuts down. It's convenient in the main thread when the executor is spinning in the background.

On a second thought, one could also disable background spinning and spin explicitly in the main thread. I'll add that option too.

PR #32, opened 2023-10-10
Extend process-wide machinery to handle more idioms

mhidalgo-bdai:
Also adds examples of use and additional tests. This PR is the result of migrating more packages downstream, which brought new use cases I had missed the first time I crafted the process-wide machinery.

### DIFF
--- a/bdai_ros2_wrappers/examples/background_action_example.py
+++ b/bdai_ros2_wrappers/examples/background_action_example.py
@@ -1,0 +1,70 @@
+# Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
+
+"""
+An example of a ROS 2 aware executable using process-wide machinery.
+
+Run with:
+
+```sh
+python3 examples/background_action_example.py
+```
+
+And follow the instructions on screen.
+"""
+
+import sys
+from typing import Any
+
+from example_interfaces.action import Fibonacci
+from rclpy.action.client import ActionClient
+from rclpy.action.server import ActionServer, ServerGoalHandle
+
+import bdai_ros2_wrappers.process as ros_process
+from bdai_ros2_wrappers.node import Node
+
+
+class MinimalActionServer(Node):
+    """A minimal ROS 2 node serving an example_interfaces.action.Fibonacci action."""
+
+    def __init__(self, node_name: str = "minimal_action_server", **kwargs: Any) -> None:
+        super().__init__(node_name, **kwargs)
+        self._action_server = ActionServer(self, Fibonacci, "compute_fibonacci_sequence", self.execute_callback)
+
+    def execute_callback(self, goal_handle: ServerGoalHandle) -> Fibonacci.Result:
+        sequence = [0, 1]
+
+        for i in range(1, goal_handle.request.order):
+            sequence.append(sequence[i] + sequence[i - 1])
+            feedback = Fibonacci.Feedback()
+            feedback.sequence = sequence
+            goal_handle.publish_feedback(feedback)
+        goal_handle.succeed()
+
+        result = Fibonacci.Result()
+        result.sequence = sequence
+        return result
+
+
+@ros_process.main()
+def main() -> None:
+    main.load(MinimalActionServer)
+
+    action_client = ActionClient(main.node, Fibonacci, "compute_fibonacci_sequence")
+    assert action_client.wait_for_server(timeout_sec=5)
+
+    while True:
+        line = input("Please provide a Fibonacci sequence order (or press Enter to exit): ")
+        if not line:
+            break
+        try:
+            order = int(line)
+        except ValueError:
+            print(f"    Hmm, '{line}' is not a number", file=sys.stderr)
+            continue
+        result = action_client.send_goal(Fibonacci.Goal(order=order))
+        print("Computed Fibonacci sequence: ", list(result.result.sequence))
+    print("Bye bye!")
+
+
+if __name__ == "__main__":
+    main()
--- a/bdai_ros2_wrappers/examples/talker_listener_example.py
+++ b/bdai_ros2_wrappers/examples/talker_listener_example.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
+
+"""
+An example of a ROS 2 multi-node process using process-wide machinery.
+
+Run with:
+
+```sh
+python3 examples/talker_listener_example.py
+```
+"""
+
+import itertools
+import typing
+
+import std_msgs.msg
+
+import bdai_ros2_wrappers.process as ros_process
+from bdai_ros2_wrappers.node import Node
+
+
+class TalkerNode(Node):
+    """A node that publishes a phrase periodically."""
+
+    def __init__(self, **kwargs: typing.Any) -> None:
+        super().__init__("talker", **kwargs)
+        self.counter = itertools.count(start=1)
+        self.pub = self.create_publisher(std_msgs.msg.String, "chat", 1)
+        rate = self.declare_parameter("rate", 1.0).value
+        self.timer = self.create_timer(1 / rate, self.callback)
+
+    def callback(self) -> None:
+        message = std_msgs.msg.String(data=f"Hi there, from {self.get_name()} (#{next(self.counter)})")
+        self.pub.publish(message)
+
+
+class ListenerNode(Node):
+    """A node that subscribes to chattering."""
+
+    def __init__(self, **kwargs: typing.Any) -> None:
+        super().__init__("listener", **kwargs)
+        self.sub = self.create_subscription(std_msgs.msg.String, "chat", self.callback, 1)
+
+    def callback(self, message: std_msgs.msg.String) -> None:
+        self.get_logger().info(message.data)
+
+
+def graph(**kwargs: typing.Any) -> typing.Iterable[Node]:
+    return [TalkerNode(**kwargs), ListenerNode(**kwargs)]
+
+
+@ros_process.main(prebaked=False, namespace=True)
+def main() -> None:
+    ros_process.spin(graph)
+
+
+if __name__ == "__main__":
+    main()
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
@@ -2,6 +2,7 @@
 import atexit
 import collections
 import concurrent.futures
+import contextlib
 import dataclasses
 import inspect
 import logging
@@ -13,7 +14,7 @@ import weakref
 
 import rclpy.executors
 
-from bdai_ros2_wrappers.utilities import fqn
+from bdai_ros2_wrappers.utilities import bind_to_thread, fqn
 
 
 class AutoScalingThreadPool(concurrent.futures.Executor):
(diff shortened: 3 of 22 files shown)


mhidalgo-bdai:
Arguably, there could be more tests here. I'm prioritizing downstream package migration.

amessing-bdai:
Can you add more comments to the examples? Some explanation about when to use what?

mhidalgo-bdai:
@amessing-bdai PTAL @ https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/1e282eb9f938fbb7f96fc8f0ee98210068b2f116 !

README.md
```diff
@@ -1,9 +1,222 @@
...
+```python
+import logging
+import time
+
+import bdai_ros2_wrappers.process as ros_process
+
+import std_msgs.msg
+
+@ros_process.main()
+def entrypoint():
```
amessing-bdai, line 27: The examples and explanations look great.

Going to ask for type hints just to keep it consistent with our code expectations.

README.md
```diff
@@ -1,9 +1,243 @@
...
+import bdai_ros2_wrappers.process as ros_process
+from bdai_ros2_wrappers.node import Node
+
+def graph(**kwargs: Any) -> List[Node]:
+    # make sure to forward all keyword arguments for proper loading!
+    return [Node("my_node", **kwargs), Node("my_other_node", **kwargs)]
+
+@ros_process.main(prebaked=False)
+def main() -> None:
+    ros_process.load(Node, "_hidden_node")  # or main.load
```
mpickett-bdai, line 72: Is this to show that we can throw in another node into the processes and that when we call spin(graph) will that then spin all 3 nodes?

mhidalgo-bdai: Exactly. And that can be done from anywhere below `main` in the call stack.

mhidalgo-bdai: It will use defaults, but you can configure it before that happens:

```python
from bdai_ros2_wrappers.executors import AutoScalingMultiThreadedExecutor
from bdai_ros2_wrappers.executors import foreground

# ...

@ros_process.main(prebaked=False)
def main() -> None:
    with foreground(AutoScalingMultiThreadedExecutor(max_threads=100)) as main.executor:
        ros_process.load(Node, "_hidden_node")  # or main.load
        ros_process.spin(graph)  # or main.spin
```

mpickett-bdai: Awesome!

README.md
```diff
@@ -1,9 +1,243 @@
...
+
+These APIs wrap [`rclpy`](https://github.com/ros2/rclpy) APIs to provide a simpler UX in common use cases.
+All APIs are thread-safe and exception-safe. Resource management is automatic. By default, callbacks are
+dispatched in the background, enabling both synchronous and asynchronous programming.
+
+These APIs are also quite handy to reconcile past [`rospy`](http://wiki.ros.org/rospy) experience with ROS 2.
+
+#### Setting up single node processes
+
+Just decorate your executable entrypoint (or `main` function) with `bdai_ros2_wrappers.process.main`:
```
jbarry-bdai, line 16:
```suggestion
To create simple single node processes that can block in common sense ways similar to ROS1's `rospy`, just decorate your executable entrypoint (or `main` function) with `bdai_ros2_wrappers.process.main`:
```

jbarry-bdai: Can you also specify what the person should see if they run this program?  I actually don't know what `create_task` does...

README.md
jbarry-bdai: Can the `README` be updated to explain that this will spin all three nodes and perhaps address a little more what `Load` does?  I had the same question reading this as @mpickett-bdai

README.md
```diff
@@ -1,9 +1,243 @@
...
+
+if __name__ == "__main__":
+    main()
+```
+
+Note the use of ROS 2 node(s) factories, to load (or spin) entire collections at once.
+
+#### Setting up interactive multi-node applications
+
+For interactive applications, you will most likely want to keep automatic spinning in place:
```
jbarry-bdai, line 83:
```suggestion
For interactive applications, you will most likely want to keep automatic spinning in place.  Here's an example that waits for user input, sleeps for 5 seconds, and then echoes that user input from a callback without deadlocking:
```

README.md
```diff
@@ -1,9 +1,243 @@
...
+def main() -> None:
+    ros_process.load(graph)  # or main.load
+    while True:
+        work(input())
+
+if __name__ == "__main__":
+    main()
+```
+
+#### Setting up command-line single node applications
```
jbarry-bdai, line 120: Could there be a sentence here about what this is showing and also what `managed` means?

README.md
```diff
@@ -1,9 +1,243 @@
...
+            root = logging.getLogger()
+            root.setLevel(logging.INFO)
+        logging.info("Application started!")
+        ros_process.spin()  # or main.spin
+
+if __name__ == "__main__":
+    main()
+```
+
+Note that process arguments are set via CLI. Also, note that a process-wide node is set explicitly (rather than implicitly).
```
jbarry-bdai, line 162: In what cases does fine grained control over ROS 2 configuration matter?  Could you expand a little here on why you want to set a process-wide node explicitly?

README.md
```diff
@@ -1,9 +1,243 @@
...
 
+## Getting started
+
+### Process-wide APIs
+
+These APIs wrap [`rclpy`](https://github.com/ros2/rclpy) APIs to provide a simpler UX in common use cases.
+All APIs are thread-safe and exception-safe. Resource management is automatic. By default, callbacks are
+dispatched in the background, enabling both synchronous and asynchronous programming.
+
+These APIs are also quite handy to reconcile past [`rospy`](http://wiki.ros.org/rospy) experience with ROS 2.
```
jbarry-bdai, line 12: I feel like a little bit more conceptual stuff here might help with understanding the examples.  I'd throw in some high level stuff about how the backend works.  For instance, a "process-wide node" is referenced a few times - I think it'd be worth saying what that is.

README.md
```diff
@@ -1,9 +1,243 @@
...
+    try:
+        time.sleep(10.0)
+    finally:
+        feature.shutdown()
+
+if __name__ == "__main__":
+    main()
+```
+
+Note the use of `ros_scope` instead of `ros_process`. Both offer the same APIs, but the former reaches out to the innermost scope. This allows the library to work in more complex, multi-threaded, multi-scope applications, and for testing cases.
```
jbarry-bdai, line 196: I'm still not clear at this point on when I'd use `ros_scope` vs `ros_process`.  Could you give an example of when scope is needed?  Is the above such an example?  What would happen if process was used instead?

README.md
```diff
@@ -1,9 +1,243 @@
...
+
+#### Writing ROS 2 aware tests
+
+To write tests, setup a global ROS 2 aware scope instead of a process.
```
jbarry-bdai, line 200:
```suggestion
To write tests, set up a global ROS 2 aware scope instead of a process.
```

README.md
```diff
@@ -1,9 +1,243 @@
...
+
+import bdai_ros2_wrappers.scope as ros_scope
+from bdai_ros2_wrappers.scope import ROSAwareScope
+
+import pytest
+
+@pytest.fixture
+def ros() -> Iterable[ROSAwareScope]:
+    with ros_scope.top(global_=True, namespace="fixture") as top:
+        yield top
```
jbarry-bdai, line 216: It's probably worth explaining what this function does for folks who aren't super up on pytest fixtures and yield.

README.md
```diff
@@ -1,9 +1,243 @@
 # ros_utilities
 Wrappers and other utilities for ROS2
 
+## Getting started
```
jbarry-bdai, line 4: I think this should also point the reader to the examples directory

bdai_ros2_wrappers/examples/background_action_example.py
```diff
@@ -0,0 +1,84 @@
...
+    When executed, an action server node is loaded, and the process-wide node is used to
+    setup an action client. This action client is then used to implement a simple console
+    application.
+    """
+    main.load(MinimalActionServer)
+
+    action_client = ActionClient(main.node, Fibonacci, "compute_fibonacci_sequence")
+    assert action_client.wait_for_server(timeout_sec=5)
+
+    while True:
```
jbarry-bdai, line 69: Should this be `True` or something that checks whether ros is still alive?  Sorry can't remember if ROS2 has that atm...

bdai_ros2_wrappers/examples/background_action_example.py
```diff
@@ -0,0 +1,84 @@
...
+        line = input("Please provide a Fibonacci sequence order (or press Enter to exit): ")
+        if not line:
+            break
+        try:
+            order = int(line)
+        except ValueError:
+            print(f"    Hmm, '{line}' is not a number", file=sys.stderr)
```
jbarry-bdai, line 76: Will this print to rosout?  If not should probably use ros logging?

bdai_ros2_wrappers/examples/talker_listener_example.py
```diff
@@ -0,0 +1,71 @@
...
+        rate = self.declare_parameter("rate", 1.0).value
+        self.timer = self.create_timer(1 / rate, self.callback)
+
+    def callback(self) -> None:
+        message = std_msgs.msg.String(data=f"Hi there, from {self.get_name()} (#{next(self.counter)})")
+        self.pub.publish(message)
+
+
+class ListenerNode(Node):
+    """A node that subscribes to chattering."""
```
jbarry-bdai, line 38:
```suggestion
    """A node that subscribes to the chat topic and echoes it."""
```

jbarry-bdai (commented):
Looked at README and examples.  I realized after putting comments in the README that a lot of them were answered in the examples, but I don't think it hurts to have the text in both places :-p

bdai_ros2_wrappers/examples/talker_listener_example.py
mhidalgo-bdai: Done in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

bdai_ros2_wrappers/examples/background_action_example.py
mhidalgo-bdai: We can also loop on context status, done in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

README.md
mhidalgo-bdai: Added a docstring in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

README.md
mhidalgo-bdai: Fixed in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

README.md
mhidalgo-bdai: Should be clearer now in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3, as the notion of scope is introduced and explained earlier.

README.md
mhidalgo-bdai: Good idea. Done in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

README.md
mhidalgo-bdai: Clarified in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

README.md
mhidalgo-bdai: Done in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/efb0db0d1ef57bdecc49aaff0dcbe9222defb1e3.

README.md
```diff
@@ -1,9 +1,277 @@
...
+    executor = ros_process.executor()
+    assert executor is not None
+    future = executor.create_task(worker)  # dispatch worker to executor
+    wait_for_future(future)  # block until worker is done
+
+@ros_process.main()
+def main() -> None:
+    ros_process.load(graph)  # or main.load
+    while True:
+        work(input())
```
amessing-bdai, line 135: Can you use `input` if this were launched via a launch file? My current understanding after digging through launch is you cannot get stdin from the launch process into a node?

mhidalgo-bdai: Right, you cannot. `launch` does not have provisions to pipe its own `stdin` to that of a process.

amessing-bdai: Ok, can mention that here? I would be concerned that someone uses `input` based on this example and then later has trouble launching.

mhidalgo-bdai: Done in https://github.com/bdaiinstitute/ros_utilities/pull/32/commits/1c1390b1712dee796beef2905f64f27a3e92b56a.

PR #35, opened 2023-10-18
Rework tf listener wrapper

mhidalgo-bdai:
This PR reworks `TFListenerWrapper` and futures' APIs to prevent the user from running into deadlocks. Builds on #32.


bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -1,116 +1,60 @@
...
-            cache_time_py = Duration(seconds=cache_time_s)
-        else:
-            cache_time_py = None
+            cache_time_py = Duration(nanoseconds=cache_time_s * 1e9)
         self._tf_buffer = Buffer(cache_time=cache_time_py)
-        self._tf_listener = TransformListener(self._tf_buffer, self._node, spin_thread=False)
-
-        if wait_for_transform is not None and len(wait_for_transform) == 2:
-            self.wait_for_init(wait_for_transform[0], wait_for_transform[1])
+        self._tf_listener = TransformListener(self._tf_buffer, self._node)
```
amessing-bdai, line 26: Is this being spun by its internal spin thread or the process-wide machinery?

amessing-bdai:
What are your thoughts about a flag (defaulting to off) for a process-wide tf listener? One that a user could access similar to the process-wide node or executor?

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
mhidalgo-bdai: A few lines above this you can see that the current scope main node is fetched if no node is explicitly provided to the wrapper. The listener is bound to either. It doesn't spin up any threads, it relies on the executor the node is (or will be) associated with.

mhidalgo-bdai:
> What are your thoughts about a flag (defaulting to off) for a process-wide tf listener? One that a user could access similar to the process-wide node or executor?

Sounds useful.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
amessing-bdai: For some reason, I thought that the TransformListener defaulted to the spin_thread being `True`.

jbarry-bdai (commented):
Only looked at tf listener wrapper

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -119,36 +63,40 @@ def lookup_a_tform_b(
...
+                buffer. If false, this will return immediately if a path does not exist even if timeout is not
+                None. Note that wait_for_a_tform_b can also be used to wait for a transform to become available.
         Returns:
             The transform frame_a_t_frame_b at the time specified.
         Raises:
             All the possible TransformExceptions.
         """
-        return self._internal_lookup_a_tform_b(frame_a, frame_b, transform_time, timeout, wait_for_frames)
+        if not wait_for_frames:
+            timeout_sec = 0.0
```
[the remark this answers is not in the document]
mhidalgo-bdai: > if the path doesn't exist, it should return immediately if wait_for_frames is true.

Hmm, that is not what the original implementation does.

> I think the way this is written it will return immediately if wait_for_frames is true and transform_time is beyond what's in the tf buffer even if the path exists from frame a to frame b.

Hmm, it shouldn't. Didn't we test this a couple weeks ago? In fact, it's because it waits that the `test_future_transform_wait` test isn't flaky due to races.

> this is useful for code that may be adding frames later in the processing and therefore if the path doesn't exist, waiting isn't going to help.

I see. It's a bit of an odd use case, as this is a problem you would run into only if you had a single-threaded, synchronous program where statements are not ordered properly. That said, we can make it behave that way. Are you sure you want it to behave that way? It will expose us to races when using lookups not guarded by waits.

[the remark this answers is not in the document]
jbarry-bdai: Wow ok I did not write this correctly:

If `wait_for_frames` is _False_ then this should return immediately if a path from frame_a to frame_b does not exist.  Otherwise it should wait for the transform to exist until the timeout (or forever if the timeout is None), which I think is the normal thing the underlying TF stuff does.

We could also potentially drop this argument as I'm not really sure anyone is using it.

bdai_ros2_wrappers/test/test_tf_listener_wrapper.py
```diff
@@ -101,7 +101,7 @@ def test_future_transform_extrapolation_exception(
     time.sleep(0.2)
     timestamp = ros.node.get_clock().now()
     with pytest.raises(ExtrapolationException):
-        tf_listener.lookup_a_tform_b(FRAME_ID, CHILD_FRAME_ID, timestamp)
+        tf_listener.lookup_a_tform_b(FRAME_ID, CHILD_FRAME_ID, timestamp, timeout_sec=0.0)
```
mhidalgo-bdai, line 104: @jbarry-bdai note this is necessary now as `None` for a timeout is always treated as infinity. Original code aliased `None` with 0 when `wait_for_frames` was `False`.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -1,154 +1,120 @@
...
 
-class TFListenerWrapper(object):
-    def __init__(
-        self,
-        node: Optional[Node] = None,
-        wait_for_transform: Optional[Tuple[str, str]] = None,
-        cache_time_s: Optional[int] = None,
-    ) -> None:
+class TFListenerWrapper:
+    def __init__(self, node: Optional[Node] = None, cache_time_s: Optional[float] = None) -> None:
```
jbarry-bdai, line 19: What happeneed to the wait for init part?  I do know people were using that...

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -1,154 +1,120 @@
...
+            wait_for_frames: If true, it will wait for a path to exist from frame_a to frame_b in the
+                buffer. If false, lookup will fail immediately if a path between frames does not exist,
+                regardless of what timeout was set. Note that wait_for_a_tform_b can also be used to
+                wait for a transform to become available.
         Returns:
             The transform frame_a_t_frame_b at the time specified.
         Raises:
             All the possible TransformExceptions.
         """
+        if not wait_for_frames:
```
jbarry-bdai, line 80: This logic now looks correct to me

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
mhidalgo-bdai: Yes, quite a bit in fact (and not always correctly e.g. [here we pass three frames](https://github.com/bdaiinstitute/bdai/blob/4e336b375747c2bcd7a20da045b165bbed46fc9e/ws/src/spot_utilities/spot_utilities/spot_grasping_server.py#L37)). It is a problematic pattern, though. When the wrapper is bound to a node that is spinning, all is fine with waiting on a constructor. When the wrapper gets instantiated from within the constructor of the node it is bound to, it hangs, because the node can't be put to spin until it's fully constructed. 

Eliminating that pattern was the original reason behind this PR. You can still wait, but you have do to so explicitly after construction.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
```diff
@@ -1,154 +1,120 @@
...
+                buffer. If false, lookup will fail immediately if a path between frames does not exist,
+                regardless of what timeout was set. Note that wait_for_a_tform_b can also be used to
+                wait for a transform to become available.
         Returns:
             The timestamp from the latest recorded transform frame_a_t_frame_b
         Raises:
             All the possible TransformExceptions.
         """
-        transform = self._internal_lookup_a_tform_b(frame_a, frame_b, None, timeout, wait_for_frames)
+        transform = self.lookup_a_tform_b(frame_a, frame_b, timeout_sec, wait_for_frames)
```
amessing-bdai, line 119: Saw `self._tf_buffer.get_latest_common_time(frame_a, frame_b)` above. Is there a difference in the result from doing that verses this?

mhidalgo-bdai: This version can wait for frames to come into existence.

mhidalgo-bdai: There is a bug on line 119 right there :eyes: Fixed it and added a test (we didn't have any for this method) in https://github.com/bdaiinstitute/ros_utilities/pull/35/commits/49ee39206451b0b9b679749494eb80106982c712.

bdai_ros2_wrappers/bdai_ros2_wrappers/tf_listener_wrapper.py
jbarry-bdai: OK.  Could we add a comment here explaining how to do that?  The reason we added this was because people were constantly getting exceptions thrown from trying to use the listener too soon.

mhidalgo-bdai: Done in https://github.com/bdaiinstitute/ros_utilities/pull/35/commits/fc71daa4b76f7c854240665a62f54527f620d797. I added documentation, including basic examples, and it will also now generate warnings if you are about to shoot yourself in the foot. PTAL!

bdai_ros2_wrappers/bdai_ros2_wrappers/futures.py
```diff
@@ -1,25 +1,27 @@
...
+        context = get_default_context()
     event = Event()
-
-    def done_callback(_: Any) -> None:
-        nonlocal event
-        event.set()
-
-    future.add_done_callback(done_callback)
-    return event.wait(timeout=timeout_sec)
+    context.on_shutdown(event.set)
```
bhung-bdai, line 24: What happens if the context has multiple events set and shuts down? Is there a possibility where two events can be waiting on each other and deadlock?

bhung-bdai (commented):
Seems fine to me, but I would want a second opinion on my concern about possible race conditions.

bdai_ros2_wrappers/bdai_ros2_wrappers/futures.py
mhidalgo-bdai: If the context has multiple events, all of them will be set. Events do not wait on each other. Callees wait on events. Callees could run into deadlock if, for example, they have to complete each other futures or you somehow manage to wait on future within an on context shutdown handler. 

So I would say deadlock are a non-issue inasmuch as it won't happen by accident unless code is flawed.

PR #39, opened 2023-10-30
Fix single-node CLI example

mhidalgo-bdai:
As reported by @dhoelscher-bdai. Current example is not only broken but it also encourages the wrong pattern.


dhoelscher-bdai (approved):
Thanks!

PR #42, opened 2023-11-21
[N/A] Clarify `prebaked` docstring

ggoretkin-bdai:


bdai_ros2_wrappers/bdai_ros2_wrappers/process.py
```diff
@@ -61,11 +61,13 @@ def __init__(
...
-            node, or, if True, the current executable basename without its extension (or CLI program
-            name if one is specified) will be used. Note that completely bare processes do not bear a
-            node nor spin an executor, which brings them closest to standard ROS 2 idioms.
+            prebaked: whether to instantiate a prebaked process or a bare process. Defaults to True.
+                - a prebaked process has an implicit node and an executor
+                - a bare process does not bear a node nor spin an executor, which brings them closest to standard
+                ROS 2 idioms.
+                - for `prebaked=True`, the ROS node name is the the current executable basename without its extension
+                (or CLI program name if one is specified)
+                - The ROS node name can be set to `foo` with `prebaked="foo"`
```
mhidalgo-bdai, line 70: @ggoretkin-bdai nit^2: structure and phrasing. 
```suggestion
            Note that:
                - a prebaked process has an implicit node and an executor;
                - a bare process has neither, which brings it the closest to standard ROS 2 idioms;
                - when `prebaked=True`, the implicit node name will be the current executable basename without extension (or CLI program name if one is specified);
                - when `prebaked="string"`, the given string will be used as the implicit node name.
```

ggoretkin-bdai:
I later noticed that there is basically the same description here: https://github.com/bdaiinstitute/ros_utilities/blob/f4b9e9d4fe8b78bcaf2548ebf40aae6834f346a0/bdai_ros2_wrappers/bdai_ros2_wrappers/scope.py#L53

mhidalgo-bdai:
> I later noticed that there is basically the same description here:

Very similar but not exactly the same.

ggoretkin-bdai:
@mhidalgo-bdai should the descriptions of `prebaked` in each case (`ROSAwareScope` and `ROSAwareProcess`) be the same? If not, perhaps we can highlight the difference.

mhidalgo-bdai:
@ggoretkin-bdai descriptions are different already, but by omission. If it helps, we can definitely mention each other and make that difference explicit.

amessing-bdai:
@ggoretkin-bdai @mhidalgo-bdai Is this PR still useful or should we close it?

mhidalgo-bdai:
@amessing-bdai @ggoretkin-bdai I think this still applies, provided the suggested docstring is in fact clearer to folks.

ggoretkin-bdai:
Could someone please rebase and merge on my behalf if it is indeed useful? Thanks.

PR #45, opened 2023-12-06
[N/A] Removed top level pyproject.toml

amessing-bdai:


mhidalgo-bdai (approved):
LGTM

I would make `configs` a hidden directory to keep only package directories at the repo root, but no biggie. It's completely subjective.

amessing-bdai:
Saw this after merging.  For now I think it is ok.

PR #46, opened 2024-01-02
Improve CI and development workflows

mhidalgo-bdai:
Split from #44, as I need the container image to be built and pushed for the actual test CI jobs to pick it up.


mhidalgo-bdai:
An argument could be made in favor of sharing container images across jobs (and runners) without pushing them. While that's possible, we would have to figure out how to do maintenance of those artifacts. Seems unwarranted at the moment.

amessing-bdai (approved):

PR #59, opened 2024-01-26
Added more ruff rules

amessing-bdai:
I've added more rules to ruff based on the ones we are adding to other repos. 

@mhidalgo-bdai Some of these involve rules for the docstring. Can you make sure that I've not misrepresented anything?

PR #58, opened 2024-01-25
Add ergonomic action API

mhidalgo-bdai:
This patch introduces Pythonic alternatives to both `ActionClientWrapper` and `ActionHandle` abstractions. Instead of clients and callbacks, we can instantiate an _actionable_. Actionables resemble remote procedure calls, and can be invoked both synchronously and asynchronously. When used asynchronously, _action futures_ replace regular futures for full access to the action lifecycle (feedback, cancellation). 

A wiki page with documentation and some examples is in order, but first I want to make sure we are aligned on the API.

PR #61, opened 2024-01-29
Ensure Python 3.8 compatibility

mhidalgo-bdai:
This slipped through during #44 and #59, and surfaced in #58. We became essentially incompatible with Python 3.8. 

This patch restores compatibility with Python 3.8.

### DIFF
--- a/proto2ros/proto2ros/utilities.py
+++ b/proto2ros/proto2ros/utilities.py
@@ -47,7 +47,7 @@ def pairwise(iterable: Iterable[Any]) -> Iterable[Tuple[Any, Any]]:
     """Yields an iterable over consecutive pairs."""
     a, b = itertools.tee(iterable)
     next(b, None)
-    return zip(a, b)  # noqa: B905
+    return zip(a, b)
 
 
 def to_ros_base_type(type_: Union[str, BaseType]) -> str:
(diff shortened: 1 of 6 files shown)


mhidalgo-bdai:
@amessing-bdai @jbarry-bdai do not merge this just yet. I want to test it against `bosdyn_msgs`.

proto2ros/proto2ros/utilities.py
```diff
@@ -47,7 +47,7 @@ def pairwise(iterable: Iterable[Any]) -> Iterable[Tuple[Any, Any]]:
     """Yields an iterable over consecutive pairs."""
     a, b = itertools.tee(iterable)
     next(b, None)
-    return zip(a, b)  # noqa: B905
```
amessing-bdai, line 50: I think you need the comment since this intentional doesn't use the strict parameter?

mhidalgo-bdai: It got removed by `ruff` or `black`. Not sure. I think it's not required in Python 3.8.

amessing-bdai: Yeah, apparently `strict` was added in 3.10...

amessing-bdai (approved):

PR #63, opened 2024-02-02
Add action future to action exceptions

mhidalgo-bdai:
As per offline discussion, propagating the action future (and thus the action result) in exceptional cases is goof first step towards action error handling. This patch does precisely that.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/action.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/action.py
@@ -21,7 +21,9 @@ from bdai_ros2_wrappers.futures import wait_for_future
 class ActionException(Exception):
     """Base action exception."""
 
-    pass
+    def __init__(self, action: "ActionFuture") -> None:
+        super().__init__(action)
+        self.action = action
 
 
 class ActionTimeout(ActionException):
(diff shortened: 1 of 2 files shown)


bdai_ros2_wrappers/bdai_ros2_wrappers/action.py
```diff
@@ -21,7 +21,9 @@
 class ActionException(Exception):
     """Base action exception."""
 
-    pass
+    def __init__(self, action: "ActionFuture") -> None:
+        super().__init__(action)
+        self.action = action
 
 
 class ActionTimeout(ActionException):
```
bhung-bdai, line 29: <i class='graphite__hidden'>[Re: lines 41 to 41]<br/><br/></i>

Should we provide a string elaborating on the error, like other exceptions do? <span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/bdaiinstitute/ros_utilities/63?utm_source=unchanged-line-comment">Graphite</a>.</span>

mhidalgo-bdai: I guess we could. I didn't because the exception types themselves are pretty self explanatory, and the action future has almost full access to action details.

amessing-bdai: I'm ok with not. There isn't a generic way to pull the status msg out of the action result and since we are providing that, the use can do it themselves

amessing-bdai (approved):

PR #68, opened 2024-02-07
Added flag to not wait for action server

amessing-bdai:
I know this will soon be replaced with Actionables, but until that happens I would like the ability to not wait for the server as its needed to add some unit tests in the main repo.


mhidalgo-bdai (approved):
LGTM

PR #84, opened 2024-03-29
Speed up `rclpy` loggers 

mhidalgo-bdai:
This patch introduces an alternative implementation of the `rclpy.logging.RcutilsLogger` that will cache log call configurations. This has been observed to reduce the average logging time significantly. 

As to why this is the case, other than the obvious advantage of executing code on first call rather than on every call, I suspect stack frame inspection in Python may be taking some amount of time (as it fetches code from source files by default when populating frame info and traceback objects, see [here](https://github.com/python/cpython/blob/fad48ea1816be3125ea51edcdfe2f999d6ade796/Lib/inspect.py#L1707)).

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
@@ -1,30 +1,37 @@
-# Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
+# Copyright (c) 2023-2024 Boston Dynamics AI Institute Inc.  All rights reserved.
 
 import contextlib
+import functools
+import inspect
 import logging
 import typing
+import warnings
 
-import rclpy.logging
+from rclpy.clock import Clock
+from rclpy.duration import Duration
 
 # NOTE(hidmic): tinkering with implementation details is not ideal,
 # but there is not a lot of room for alternatives. This inconvenience
 # will go away when and if this module is contributed upstream.
 from rclpy.impl.implementation_singleton import rclpy_implementation as impl
+from rclpy.impl.rcutils_logger import RcutilsLogger
+from rclpy.logging import LoggingSeverity
 from rclpy.node import Node
 
+from bdai_ros2_wrappers.utilities import cap, skip, throttle
+
 SEVERITY_MAP = {
-    logging.NOTSET: rclpy.logging.LoggingSeverity.UNSET,
-    logging.DEBUG: rclpy.logging.LoggingSeverity.DEBUG,
-    logging.INFO: rclpy.logging.LoggingSeverity.INFO,
-    logging.WARN: rclpy.logging.LoggingSeverity.WARN,
-    logging.ERROR: rclpy.logging.LoggingSeverity.ERROR,
-    logging.CRITICAL: rclpy.logging.LoggingSeverity.FATAL,
+    logging.NOTSET: LoggingSeverity.UNSET,
+    logging.DEBUG: LoggingSeverity.DEBUG,
+    logging.INFO: LoggingSeverity.INFO,
+    logging.WARN: LoggingSeverity.WARN,
+    logging.ERROR: LoggingSeverity.ERROR,
+    logging.CRITICAL: LoggingSeverity.FATAL,
 }
 
 
 if not typing.TYPE_CHECKING:
     import os.path
-    import warnings
     import xml.etree.ElementTree as ET
 
     import ament_index_python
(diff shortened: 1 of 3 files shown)


robodreamer:
Thanks for proposing a quick alternative solution! Would you mind elaborating on what the issues in the original code were and how the new change addresses them?

mhidalgo-bdai:
> Thanks for proposing a quick alternative solution! Would you mind elaborating on what the issues in the original code were and how the new change addresses them?

I updated the PR description. I think it's a combination of factors that can be summarized as "too much work on every log call". This changes addresses this by "doing less work once".

bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
```diff
@@ -37,6 +44,229 @@
...
 
+@functools.lru_cache(maxsize=1024)
+def make_logging_function(
+    name: str,
+    level: LoggingSeverity,
+    origin: inspect.Traceback,
+    throttle_duration_sec: typing.Optional[float] = None,
+    throttle_time_source: typing.Optional[Clock] = None,
+    skip_first: typing.Optional[bool] = None,
+    once: typing.Optional[bool] = None,
```
amessing-bdai, line 55: What is the difference between `None` and `False`?

mhidalgo-bdai: In this context, none. I annotate optional arguments as optional purely out of habit (and for the sake of readability).

bdai_ros2_wrappers/test/test_logging.py
```diff
@@ -1,15 +1,80 @@
 # Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
 import logging
-from typing import Optional
+import threading
+from typing import List, Optional
```
robodreamer, line 4: we could use `list` and not import this.

bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
```diff
@@ -37,6 +44,229 @@
...
+
+    def is_enabled_for(self, level: typing.Union[int, LoggingSeverity]) -> bool:
+        """Checks whether the logger is enabled for logs at the given severity `level`."""
+        return impl.rclpy_logging_logger_is_enabled_for(self.name, LoggingSeverity(level))
+
+    def log(
+        self,
+        message: str,
+        level: typing.Union[int, LoggingSeverity],
+        origin: typing.Optional[inspect.Traceback] = None,
```
robodreamer, line 149: `type | None` could be used in replacement of `Optional`

robodreamer (approved):
Thanks for putting a quick workaround! I wonder if we would like to contribute this change back to ROS2 as well?

robodreamer:
I am looking forward to seeing this change live from the bdai container soon! I had to suppress a lot of log messages currently to get around the performance issue.

bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
```diff
@@ -37,6 +44,229 @@
...
+            fill_value=False,
+        )
+    if skip_first:
+        log = skip(log, num_times=1, fill_value=False)
+    if once:
+        log = cap(log, num_times=1, fill_value=False)
+    return log
+
+
+class MemoizingRcutilsLogger:
```
robodreamer, line 109: This this class name a standard one? The name (`memoizing` sounds a bit new to me.

bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
amessing-bdai: @robodreamer `ros_utilities` supports python 3.8 which doesn't support the pipe operator for types.

bdai_ros2_wrappers/test/test_logging.py
amessing-bdai: Same as above about 3.8

bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
amessing-bdai: Ok, not going to block on this.

I think whether `bool = False` or `Optional[bool] = None` is more readable is another discussion lol.

bdai_ros2_wrappers/bdai_ros2_wrappers/logging.py
mhidalgo-bdai: Hmm, not standard, no. Memoizing is me adjectivizing [memoization](https://en.wikipedia.org/wiki/Memoization).

mhidalgo-bdai:
> I wonder if we would like to contribute this change back to ROS2 as well?

I'm onboard.

> I am looking forward to seeing this change live from the bdai container soon! I had to suppress a lot of log messages currently to get around the performance issue.

Note you will have to wrap your logger, like `logger = as_memoizing_logger(node.get_logger())`. We could inject it for our `bdai_ros2_wrappers.node.Node` subclass, but I suspect that may create some confusion when people start mixing up our nodes with regular nodes (and ask why the logger in their regular node has awful performance).

PR #85, opened 2024-04-03
[SW-922] Add wait_for_messages to ros utilities

kaiyu-zheng:
Add wait_for_messages function from genmos/ros2_utils to [subscription.py](https://github.com/bdaiinstitute/ros_utilities/blob/main/bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py),  and unit tested.


bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,137 @@ def wait_for_message(
...
+        if latched_topics is None:
+            latched_topics = set()
+        self.latched_topics = latched_topics
+        self.logger = rclpy.impl.rcutils_logger.RcutilsLogger(name="wait_for_messages")
+
+        if self.verbose:
+            self.logger.info("initializing message filter ApproximateTimeSynchronizer")
+        self.subs: typing.Optional[typing.List] = [
+            self._message_filters_subscriber(mtype, topic, callback_group=callback_group)
+            for topic, mtype in zip(topics, mtypes, strict=True)  # type: ignore
```
kaiyu-zheng, line 154: Note on `type: ignore`. I got this mypy error and didn't know how to resolve it
```
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: error: No overload variant of "zip" matches argument types "List[Any]", "List[Any]", "bool"  [call-overload]
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note: Possible overload variants:
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note:     def [_T_co, _T1] zip(cls, Iterable[_T1], /) -> zip[Tuple[_T1]]
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note:     def [_T_co, _T1, _T2] zip(cls, Iterable[_T1], Iterable[_T2], /) -> zip[Tuple[_T1, _T2]]
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note:     def [_T_co, _T1, _T2, _T3] zip(cls, Iterable[_T1], Iterable[_T2], Iterable[_T3], /) -> zip[Tuple[_T1, _T2, _T3]]
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note:     def [_T_co, _T1, _T2, _T3, _T4] zip(cls, Iterable[_T1], Iterable[_T2], Iterable[_T3], Iterable[_T4], /) -> zip[T
uple[_T1, _T2, _T3, _T4]]
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note:     def [_T_co, _T1, _T2, _T3, _T4, _T5] zip(cls, Iterable[_T1], Iterable[_T2], Iterable[_T3], Iterable[_T4], Iterab
le[_T5], /) -> zip[Tuple[_T1, _T2, _T3, _T4, _T5]]
bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py:157: note:     def [_T_co] zip(cls, Iterable[Any], Iterable[Any], Iterable[Any], Iterable[Any], Iterable[Any], Iterable[Any], /
, *iterables: Iterable[Any]) -> zip[Tuple[Any, ...]]
```

bdai_ros2_wrappers/test/test_wait_for_messages.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2023 Boston Dynamics AI Institute LLC. All rights reserved.
```
mhidalgo-bdai, line 1: @kaiyu-zheng nit^2: could we relocate this to `test_subscription.py` for consistency?

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
mhidalgo-bdai: I suspect differences in Python versions for `mypy`. The `strict` argument for `zip` was introduced in 3.10.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,137 @@ def wait_for_message(
...
+        topics: typing.List,
+        mtypes: typing.List,
+        queue_size: int = 10,
+        delay: float = 0.2,
+        allow_headerless: bool = False,
+        sleep: float = 0.5,
+        timeout: typing.Optional[float] = None,
+        verbose: bool = False,
+        exception_on_timeout: bool = False,
+        latched_topics: typing.Optional[typing.Set] = None,
```
mhidalgo-bdai, line 135: @kaiyu-zheng meta: there is no such thing as a latched topic in ROS 2. Consider allowing the user to specify QoS policies for each topic. Perhaps a list of (type, name, qos) triplets, like `create_subscription` takes?

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,137 @@ def wait_for_message(
...
+
+        try:
+            self._start_time = self.node.get_clock().now()
+            rate = self.node.create_rate(1.0 / sleep)
+            while self.messages is None and not self.has_timed_out:
+                if self.check_messages_received():
+                    break
+                rate.sleep()
+        finally:
+            self.destroy_subs()
```
mhidalgo-bdai, line 172: @kaiyu-zheng meta: by turning this into a `future`, resolved by `_cb`, you can decouple and defer much of the wait logic (incl. timeouts and exceptions) to the user. No need to do expensive polling either. See `wait_for_message_async` for reference on how a future can be instantiated and resolved. Also on how to tie subscription cleanup to future resolution.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,137 @@ def wait_for_message(
         future.cancel()
         return None
     return future.result()
+
+
+def wait_for_messages(
+    node: rclpy.node.Node,
```
mhidalgo-bdai, line 87: @kaiyu-zheng nit: consider making `node` optional, fetching it from the outer `scope` by default like the other functions in this module do.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
mhidalgo-bdai: Once the set of arguments gets reduced, you can just forward `**kwargs` to `message_filters.ApproximateTimeSynchronizer`.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,137 @@ def wait_for_message(
...
+        self.messages: typing.Optional[typing.Tuple] = None
+        self.verbose = verbose
+        self.topics = topics
+        self.timeout = timeout
+        self.exception_on_timeout = exception_on_timeout
+        self.has_timed_out = False
+        if latched_topics is None:
+            latched_topics = set()
+        self.latched_topics = latched_topics
+        self.logger = rclpy.impl.rcutils_logger.RcutilsLogger(name="wait_for_messages")
```
mhidalgo-bdai, line 148: @kaiyu-zheng consider using the node logger instead.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,137 @@ def wait_for_message(
...
+                each topic.
+            latched_topics (set): a set of topics for which the publisher latches (i.e.
+                sets QoS durability to transient_local).
+
+    """
+    return _WaitForMessages(node, topics, mtypes, **kwargs).messages
+
+
+class _WaitForMessages:
+    def __init__(
```
mhidalgo-bdai, line 123: @kaiyu-zheng meta: using `__init__` methods to do work is a fairly unusual idiom. If you want to bundle the logic in an object for reuse (rather than having a temporary instantiated on each `wait_for_messages` invocation), consider relocating to a e.g. `wait_for_messages()` method.

mhidalgo-bdai (commented):
First pass!

bdai_ros2_wrappers/test/test_wait_for_messages.py
kaiyu-zheng: OK, I can do that. I thought about it at first but was hesitant because there are two node classes created specifically for this test, not for other tests in `test_subscription.py`. What's your take on this concern?

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: I thought it may be a convenient feature to provide a default QoS profile setting for the latching use case (although latch is not a concept in ROS2). But I understand this is not standard practice for ROS2. Can make the change.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: By user, do you mean the user who called "wait_for_messages"? Why would the user want to handle this logic? My take is from the user's perspective, they just want to wait till messages are returned. 

Do you think making this change would affect how the test is written? I personally like the way `wait_for_messages` is used from the user perspective as shown in the test.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: I agree that this is unusual.  I could refactor so that the logic in `__init__` is in another method in `_WaitForMessages`. If we do the `futures` route as proposed above, then that would affect this refactoring I suppose. But that also influences user experience (and the test). I think we should first agree about that.

kaiyu-zheng:
Thanks @mhidalgo-bdai !

I think we should find agreement about the test behavior / user-experience first. Quoting the comment regarding using futures:
>Do you think making this change would affect how the test is written? I personally like the way wait_for_messages is used from the user perspective as shown in the test.

bdai_ros2_wrappers/test/test_wait_for_messages.py
kaiyu-zheng: Done.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: Addressed the first comment.

Regarding `**kwargs`, the kwargs for `message_filters.Subscriber` and `message_filters.ApproximateTimeSynchronizer` are different, so I can't pass `**kwargs` into `message_filters.ApproximateTimeSynchronizer`'s constructor. Lmk if I'm missing something.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: This is great advice. It simplified the code so much, and doesn't really affect the user interface. 

Refactored. Thanks!!

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: Again thanks for the advice. Refactored!

(Ignore my comment above)

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: I removed `strict=True` and mypy is happy now.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: Done.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
kaiyu-zheng: There is no logger any more, similar to `wait_for_message`. 

It would be nice if `wait_for_future` in `futures.py` can (optionally) log a message periodically, indicating that it's waiting. But this is out of scope for this PR.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,106 @@ def wait_for_message(
...
+        queue_size,
+        delay,
+        allow_headerless=allow_headerless,
+    )
+    ts.registerCallback(callback)
+    future.add_done_callback(lambda future: _destroy_subs(node, subs))
+    return future
+
+
+def _destroy_subs(node: rclpy.node.Node, subs: typing.List[message_filters.Subscriber]) -> None:
```
mhidalgo-bdai, line 183: @kaiyu-zheng FYI: you can define this done callback in `wait_for_messages_async` scope like you did for `callback`.

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,106 @@ def wait_for_message(
...
+    topics: typing.List,
+    mtypes: typing.List,
+    *,
+    node: typing.Optional[rclpy.node.Node] = None,
+    timeout_sec: typing.Optional[float] = None,
+    **kwargs: typing.Any,
+) -> typing.Union[None, typing.List[MessageT]]:
+    """Waits for messages to arrive at multiple topics within a given time window.
+
+    Uses message_filters.ApproximateTimeSynchronizer. This function blocks
```
mhidalgo-bdai, line 96: @kaiyu-zheng meta: would it make to allow the user to pass its own synchronizer classes? Furthermore, we could define function arguments such that `**kwargs` can be forwarded to the synchronizer constructor directly.

bdai_ros2_wrappers/test/test_subscription.py
```diff
@@ -22,3 +25,64 @@ def deferred_publish() -> None:
...
+            self.pub = self.create_publisher(std_msgs.msg.String, "/test2", 10)
+            self.timer = self.create_timer(0.5, self.publish_msg)
+
+        def publish_msg(self) -> None:
+            self.pub.publish(std_msgs.msg.String(data="hello from bar"))
+
+    ros.load(NodeFoo)
+    ros.load(NodeBar)
+
+    node_wfm = ros.node
```
mhidalgo-bdai, line 52: @kaiyu-zheng nit: you don't need to pass the node explicitly, it's in scope.

bdai_ros2_wrappers/test/test_subscription.py
```diff
@@ -22,3 +25,64 @@ def deferred_publish() -> None:
...
+    class NodeFoo(Node):
+        def __init__(self, **kwargs: Any) -> None:
+            super().__init__("foo", **kwargs)
+            self.pub = self.create_publisher(std_msgs.msg.String, "/test1", 10)
+            self.timer = self.create_timer(0.5, self.publish_msg)
+
+        def publish_msg(self) -> None:
+            self.pub.publish(std_msgs.msg.String(data="hello from foo"))
+
+    class NodeBar(Node):
```
mhidalgo-bdai, line 40: @kaiyu-zheng nit^2: you could reuse the same node class and make it configurable (hint: you can `ros.load(NodeClass, arg1, arg2, kwarg1=..., kwarg2=...)`)

bdai_ros2_wrappers/bdai_ros2_wrappers/subscription.py
```diff
@@ -78,3 +81,106 @@ def wait_for_message(
...
+
+
+def wait_for_messages_async(
+    topics: typing.List,
+    mtypes: typing.List,
+    *,
+    queue_size: int = 10,
+    delay: float = 0.2,
+    allow_headerless: bool = False,
+    sleep: float = 0.5,
```
mhidalgo-bdai, line 133: @kaiyu-zheng unused?

mhidalgo-bdai (commented):
Second pass. Great work @kaiyu-zheng !

mhidalgo-bdai:
Rebased to solve conflicts, then updated to match review comments. @amessing-bdai @kaiyu-zheng PTAL.

kaiyu-zheng:
LGTM. Looks like the comment that hasn't been addressed is allowing user to pass in custom synchronizer. I am happy with leaving this for the next PR and have ApproximateTimeSynchronizer there by default.

mhidalgo-bdai:
> Looks like the comment that hasn't been addressed is allowing user to pass in custom synchronizer.

I gave it some more thought and ended up deciding against. We cannot specify sane default arguments for the synchronizer if we don't know which synchronizer (well, you can, there is always a way, but the code gets uglier).

mhidalgo-bdai:
@amessing-bdai @jbarry-bdai @bhung-bdai PTAL

bhung-bdai (approved):

PR #88, opened 2024-04-11
Fix `wait_for_future` for cancelled futures

mhidalgo-bdai:
This patch copes with how `rclpy.task.Future` deals with cancellation (done callbacks trigger on cancel but the future never becomes done, a done callback added to a cancelled future will never be invoked because the future is not done 😵‍💫).


amessing-bdai:
@dyershov-bdai Is this what you were talking about the other day?

amessing-bdai (approved):

PR #89, opened 2024-04-22
[EK-35] Use `MemoizingRcutilsLogger` in `bdai_ros2_wrappers.node.Node`

mhidalgo-bdai:
Follow-up to #84. In hindsight, automatically providing a memoizing logger to all `bdai_ros2_wrappers.node.Node` users makes for a much less painful migration and it spares us two migration cycles when and if we submit this upstream.


amessing-bdai (approved):

amessing-bdai:
@mhidalgo-bdai Lets just make sure to announce this when this hits bdai

robodreamer (approved):
Nice find! Thanks!

PR #91, opened 2024-05-03
Add `nothrow` option to ergonomic wrappers

mhidalgo-bdai:
This patch adds the `nothrow` option to synchronous calls to services and actions through existing wrappers. This simplifies certain workflows (such as complying with legacy API contracts).

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/action.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/action.py
@@ -142,6 +142,10 @@ class ActionFuture:
         Raises:
             RuntimeError: if action is still executing.
         """
+        if not self.acknowledged:
+            raise RuntimeError("Action not acknowledged")
+        if not self.accepted:
+            raise RuntimeError("Action not accepted")
         if self._result_future is None or not self._result_future.done():
             raise RuntimeError("Action still executing")
         return self._result_future.result().result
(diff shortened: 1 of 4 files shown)


bdai_ros2_wrappers/bdai_ros2_wrappers/action.py
```diff
@@ -379,6 +383,7 @@ def synchronously(
         *,
         feedback_callback: Optional[Callable] = None,
         timeout_sec: Optional[float] = None,
+        nothrow: bool = False,
```
amessing-bdai, line 386: Should `nothrow` be set here or in the constructor? The `result` property also has the chance of throwing

mhidalgo-bdai: > Should nothrow be set here or in the constructor?

It could be set on construction, though I'd rather not encourage the workflow. This is a escape hatch for existing code.

> The `result` property also has the chance of throwing

Yes. Same with the constructor. Same for `wait_for_future`. A strict `nothrow` would require a try-catch-all clause. This simply skips the nominal exceptions. We can make it strict. Existing APIs are not really explicit about their failure modes.

amessing-bdai: If this fits the need then I think it is fine. No need to add extra infrastructure if we don't need it

PR #97, opened 2024-05-29
Deduplicate scope node checks

mhidalgo-bdai:
Precisely what the title says. Too many instances of these 2-3 lines, and more to come. This patch deduplicates them all with `ensure_node()`.


khughes-bdai (approved):
LGTM

PR #98, opened 2024-05-29
Add `TransformFilter`

mhidalgo-bdai:
Python equivalent to the C++ [`tf2_ros.MessageFilter`](https://github.com/ros2/geometry2/blob/ad72a82961e82470d2c9848840dd6e8eee44b127/tf2_ros/include/tf2_ros/message_filter.h#L120) that is available upstream.


khughes-bdai (approved):
LGTM

PR #103, opened 2024-06-10
Add utilities to deal with time and duration

mhidalgo-bdai:
Precisely what the title says. Use of Time and Duration representations in Python, `rclpy`, `rosbag2`, and our own codebase is not particularly homogeneous. This patch introduces minimal conventions to simplify API implementation.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
@@ -8,7 +8,12 @@ from rclpy.time import Time
 
 
 def as_proper_time(time: Union[int, float, datetime, Time]) -> Time:
-    """Return `time` as a proper Time object."""
+    """Return `time` as a proper Time object.
+
+    Note that for scalar times the convention is that floating point times
+    are expressed in seconds since some clock epoch, and integral times are
+    expressed in nanoseconds since some clock epoch.
+    """
     if isinstance(time, int):
         return Time(nanoseconds=time)
     if isinstance(time, float):
@@ -19,7 +24,11 @@ def as_proper_time(time: Union[int, float, datetime, Time]) -> Time:
 
 
 def as_proper_duration(duration: Union[int, float, timedelta, Duration]) -> Duration:
-    """Return `duration` as a proper Duration object."""
+    """Return `duration` as a proper Duration object.
+
+    Note that for scalar durations the convention is that floating point durations
+    are expressed in seconds, and integral durations are expressed in nanoseconds.
+    """
     if isinstance(duration, int):
         return Duration(nanoseconds=duration)
     if isinstance(duration, float):
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
@@ -1,0 +1,29 @@
+# Copyright (c) 2024 Boston Dynamics AI Institute Inc.  All rights reserved.
+
+from datetime import datetime, timedelta
+from typing import Union
+
+from rclpy.duration import Duration
+from rclpy.time import Time
+
+
+def as_proper_time(time: Union[int, float, datetime, Time]) -> Time:
+    """Return `time` as a proper Time object."""
+    if isinstance(time, int):
+        return Time(nanoseconds=time)
+    if isinstance(time, float):
+        return Time(seconds=time)
+    if isinstance(time, datetime):
+        return Time(seconds=time.timestamp())
+    return time
+
+
+def as_proper_duration(duration: Union[int, float, timedelta, Duration]) -> Duration:
+    """Return `duration` as a proper Duration object."""
+    if isinstance(duration, int):
+        return Duration(nanoseconds=duration)
+    if isinstance(duration, float):
+        return Duration(seconds=duration)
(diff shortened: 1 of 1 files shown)


bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
```diff
@@ -0,0 +1,29 @@
...
+from rclpy.duration import Duration
+from rclpy.time import Time
+
+
+def as_proper_time(time: Union[int, float, datetime, Time]) -> Time:
+    """Return `time` as a proper Time object."""
+    if isinstance(time, int):
+        return Time(nanoseconds=time)
+    if isinstance(time, float):
+        return Time(seconds=time)
```
khughes-bdai, line 15: I feel that this float/int discrepancy could lead to confusion. I could potentially see someone trying to call this function to represent 5 seconds by passing in 5 instead of 5.0 and instead getting a representation that corresponds to 5 ns. Is there a way we can make this distinction more clear?

mhidalgo-bdai: That is a possibility, yes. The fool proof solution is to stop taking scalars for time/duration and force users to use explicit representations. That makes it somewhat inconvenient when you add `numpy`, `scipy`, `pandas`, etc. to the mix, but you can't go wrong that way. The second best solution is to assume SI units for scalars -- won't save you from inadvertently passing in a nanosecond count.

khughes-bdai: Totally agree with that -- I guess I'm wondering if there a clean way to specify what type corresponds to what unit in the docstring to try to mitigate this?

khughes-bdai (approved):
LGTM -- just left a minor comment

bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
```diff
@@ -0,0 +1,38 @@
...
+    if isinstance(time, datetime):
+        return Time(seconds=time.timestamp())
+    return time
+
+
+def as_proper_duration(duration: Union[int, float, timedelta, Duration]) -> Duration:
+    """Return `duration` as a proper Duration object.
+
+    Note that for scalar durations the convention is that floating point durations
```
robodreamer, line 29: Do we want to add a proper docstring with input specification including units?

robodreamer (approved):
Left a minor comment, but looks nice.

bdai_ros2_wrappers/bdai_ros2_wrappers/time.py
mhidalgo-bdai: Hmm, I'm not exactly sure what you mean by _input specification_.

robodreamer: I meant, something like the following:

```suggestion
    """
    Convert various duration inputs into a standardized Duration object.

    Parameters:
    duration (Union[int, float, timedelta, Duration]): The duration to be converted.
        - If an int is provided, it is assumed to be the duration in seconds.
        - If a float is provided, it is assumed to be the duration in seconds.
        - If a timedelta is provided, it is converted to the appropriate Duration object.
        - If a Duration object is provided, it is returned as is.

    Returns:
    Duration: A standardized Duration object representing the input duration.

    Raises:
    ValueError: If the input duration is not of a supported type.

    Example:
    >>> as_proper_duration(3600)
    <Duration: 3600 seconds>
    >>> as_proper_duration(2.5)
    <Duration: 2.5 seconds>
    >>> as_proper_duration(timedelta(hours=1))
    <Duration: 3600 seconds>
    >>> as_proper_duration(Duration(3600))
    <Duration: 3600 seconds>
    """
```

robodreamer: I meant, something like the following:

```
    """
    Convert various duration inputs into a standardized Duration object.

    Parameters:
    duration (Union[int, float, timedelta, Duration]): The duration to be converted.
        - If an int is provided, it is assumed to be the duration in seconds.
        - If a float is provided, it is assumed to be the duration in seconds.
        - If a timedelta is provided, it is converted to the appropriate Duration object.
        - If a Duration object is provided, it is returned as is.

    Returns:
    Duration: A standardized Duration object representing the input duration.

    Raises:
    ValueError: If the input duration is not of a supported type.

    Example:
    >>> as_proper_duration(3600)
    <Duration: 3600 seconds>
    >>> as_proper_duration(2.5)
    <Duration: 2.5 seconds>
    >>> as_proper_duration(timedelta(hours=1))
    <Duration: 3600 seconds>
    >>> as_proper_duration(Duration(3600))
    <Duration: 3600 seconds>
    """
```

mhidalgo-bdai: See https://github.com/bdaiinstitute/ros_utilities/pull/103/commits/c3262c8e1f3c5091d6420d5c1450352e90a5dfe4. I purposely opted out from duplicating type annotations in docstrings.

mhidalgo-bdai:
It looks like #96 wasn't enough. We keep getting flakes from the log forwarding test -.-

amessing-bdai (approved):

robodreamer (approved):
Thanks for addressing the comments!

PR #105, opened 2024-06-10
Add support for adapted message feeds

mhidalgo-bdai:
This patch allows for adapter patterns in message feeds.


khughes-bdai (approved):
LGTM

PR #109, opened 2024-07-26
Add callbacks to message feeds

mhidalgo-bdai:
In some cases, a callback-based interface is still handy. This patch supports the corresponding idioms.


amessing-bdai (approved):

khughes-bdai (approved):
LGTM

PR #111, opened 2024-08-12
Allow generalized method documentation

mhidalgo-bdai:
This patch prepares the ground for proper documentation and type annotation of generalized methods. Specifically, it decouples method prototype and legacy implementation, and it enables protocol-based type annotations for these relatively complex devices.


tcappellari-bdai (approved):
lgtm

PR #112, opened 2024-08-12
Add py.typed marker

mhidalgo-bdai:
Precisely what the title says. This patch installs the `py.typed` marker along with `bdai_ros2_wrappers`, and fixes typing annotations that start to fail.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
@@ -1,12 +1,11 @@
 # Copyright (c) 2023 Boston Dynamics AI Institute Inc.  All rights reserved.
-from typing import Callable, Optional, Type
+from typing import Any, Callable, Optional, Type
 
 import rclpy.action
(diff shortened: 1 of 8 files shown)


bdai_ros2_wrappers/bdai_ros2_wrappers/action_client.py
```diff
@@ -1,28 +1,27 @@
...
-from bdai_ros2_wrappers.type_hints import Action
 
 
 class ActionClientWrapper(rclpy.action.ActionClient):
     """A wrapper for ros2's ActionClient for extra functionality"""
 
     def __init__(
         self,
-        action_type: Type[Action],
+        action_type: Type,
```
mhidalgo-bdai, line 16: Note for reviewers: this patch drops typing annotations for legacy client and handle wrappers, as the existing ones are not properly setup (ie. they don't work the way they were intended to work) and making the necessary changes would require lots of changes downstream that we may as well skip as we move on from these.

amessing-bdai (approved):

PR #118, opened 2024-09-13
Add C++ conversion APIs to `proto2ros`

mhidalgo-bdai:
Precisely what the title says. We are almost there, I'm just missing the C++ test suite but conversion code compiles and looks correct on visual inspection. We can start the review.


proto2ros/proto2ros/output/templates/conversions.cpp.jinja
```diff
@@ -0,0 +1,396 @@
...
+        {%- set type_spec = known_message_specifications.get(to_ros_base_type(spec.type)) -%}
+        {%- if type_spec and type_spec.annotations.get("proto-class") == "enum" -%}
+            {#- Handle enum case i.e. extract integral values from ROS wrapper messages. -#}
+            {%- if spec.type.is_array -%}
+{{ destination }}->Reserve({{ source }}.size());
+for (const auto& item : {{ source }}) {
+    {{ destination }}->Add(static_cast<{{ type_spec.annotations["proto-type"] | as_pb2_cpp_type }}>(item.value));
+}
+            {%- else -%}
+{{ destination | rreplace("mutable", "set", 1) | rreplace("()", "", 1) }}(
```
mhidalgo-bdai, line 37: I'm not super proud of this but the C++ API for Protobuf message is so inconsistent across types that it hurts.

proto2ros/proto2ros/output/templates/conversions.cpp.jinja
```diff
@@ -0,0 +1,396 @@
...
+   static_cast<{{ type_spec.annotations["proto-type"] | as_pb2_cpp_type }}>({{ source }}.value));
+            {%- endif -%}
+        {%- elif spec.type.is_array -%}
+            {%- set input_item = itemize_cpp_identifier(source, "input_") -%}
+            {%- set output_item = itemize_cpp_identifier(destination, "output_") -%}
+            {%- if type_spec and type_spec.annotations.get("map-entry") -%}{#- Handle map case. -#}
+{{ destination }}->clear();
+for (const auto& {{ input_item }} : {{ source }}) {
```
mhidalgo-bdai, line 45: Handling map fields was particularly hard. Python APIs for Protobuf messages expose the underlying sequence of key-value messages. C++ APIs hide them, and that forces special casing :facepalm:

proto2ros_tests/include/proto2ros_tests/manual_conversions.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <geometry_msgs/msg/twist.hpp>
+#include <geometry_msgs/msg/vector3.hpp>
+#include <geometry_msgs/msg/wrench.hpp>
+
+#include <sensor_msgs/msg/temperature.hpp>
+
+namespace proto2ros_tests {
+namespace conversions {
+
+void convert(const geometry_msgs::msg::Vector3& ros_msg, bosdyn::api::Vec3* proto_msg);
```
mhidalgo-bdai, line 20: Note for myself: this is missing documentation.

proto2ros/cmake/proto2ros_vendor_package.cmake
```diff
@@ -55,4 +56,28 @@ macro(proto2ros_vendor_package target)
...
+  ament_export_include_directories("include/${PROJECT_NAME}")
+
+  install(
+    TARGETS ${target}_conversions
+    EXPORT export_${target}_conversions
+    ARCHIVE DESTINATION lib
+    LIBRARY DESTINATION lib
+    RUNTIME DESTINATION bin
+  )
+  ament_export_targets(export_${target}_conversions)
```
mhidalgo-bdai, line 82: Note for myself: this is likely incomplete, and will need a revisit once I try to re-build https://github.com/bdaiinstitute/bosdyn_msgs with this.

proto2ros/include/proto2ros/conversions.hpp
```diff
@@ -0,0 +1,154 @@
...
+#include <proto2ros/msg/value.hpp>
+
+namespace proto2ros {
+namespace conversions {
+
+/// Unpacks a proto2ros/AnyProto ROS message into any Protobuf message.
+///
+/// \throws std::runtime_error if the given ROS message cannot be unpacked onto the given Protobuf message.
+template <typename T>
+void convert(const proto2ros::msg::AnyProto& ros_msg, T* proto_msg) {
```
amessing-bdai, line 40: Thoughts about using references instead of pointers? Unless you plan to check for null everywhere I think they are safer.

amessing-bdai: Ah... yeah with `mutable` fields returning pointers instead of references I think this is fine. I don't like the `OutputT convert(const InputT&)` style because that triggers a copy/move instead of using the existing field.

.pre-commit-config.yaml
```diff
@@ -28,6 +28,16 @@ repos:
...
+- repo: https://github.com/ssciwr/clang-format-hook.git
+  rev: v14.0.0
+  hooks:
+  -   id: clang-format
+      types_or: [c++, c]
+- repo: https://github.com/cpplint/cpplint.git
+  rev: 1.6.1
+  hooks:
+  -   id: cpplint
+      args: ['--quiet', '--filter=-whitespace/comments', '--linelength=120']
```
amessing-bdai, line 40: How do you feel about `clang-tidy` as a unit test instead of `cpplint` as a `pre-commit` check?

proto2ros/CMakeLists.txt
```diff
@@ -13,19 +13,47 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_python REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
+find_package(builtin_interfaces REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(rclcpp REQUIRED)
+
+find_package(Protobuf REQUIRED)
```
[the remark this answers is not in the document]
amessing-bdai:
```
 # If no build type is specified, then the build type is set to "Release"
    set(DEFAULT_BUILD_TYPE "Release")
    if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
        message(STATUS "Setting build type to '${DEFAULT_BUILD_TYPE}' as none was specified.")
        set(CMAKE_BUILD_TYPE "${DEFAULT_BUILD_TYPE}" CACHE STRING "Choose the type of the build." FORCE)
        # Set the possible values of the build type for cmake-gui
        set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
    endif(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
```

maybe ^ too?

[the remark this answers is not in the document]
amessing-bdai:
```
if (CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
        include(CheckIPOSupported)
        check_ipo_supported(RESULT ipo_supported OUTPUT output)
        if (ipo_supported)
            set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
            message(STATUS "We are in release - Successfully enabled IPO")
        else ()
            message(WARNING "IPO NOT SUPPORTED - Skipping reason: ${output}")
        endif ()
    endif ()
```

proto2ros/CMakeLists.txt
```diff
@@ -13,19 +13,47 @@ endif()
...
 rosidl_generate_interfaces(${PROJECT_NAME}
   msg/Any.msg msg/AnyProto.msg msg/Bytes.msg msg/List.msg
   msg/Struct.msg msg/StructEntry.msg msg/Value.msg
 )
 
+add_library(${PROJECT_NAME}_conversions SHARED src/conversions.cpp)
+target_include_directories(${PROJECT_NAME}_conversions PUBLIC
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>"
+)
```
amessing-bdai, line 31:
```
target_compile_options(${PROJECT_NAME}_conversions
    PUBLIC
    "$<$<CONFIG:DEBUG>:-O0>"
    "$<$<OR:$<CONFIG:RELWITHDEBINFO>,$<CONFIG:RELEASE>>:-O3>"
    "$<$<OR:$<CONFIG:RELWITHDEBINFO>,$<CONFIG:DEBUG>>:-g>")
target_compile_definitions(${PROJECT_NAME}_conversions
    PUBLIC
    "$<$<CONFIG:DEBUG>:DEBUG>"
    "$<$<OR:$<CONFIG:RELWITHDEBINFO>,$<CONFIG:RELEASE>>:NDEBUG>")

proto2ros/CMakeLists.txt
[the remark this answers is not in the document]
mhidalgo-bdai: >             set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)

TIL

proto2ros/CMakeLists.txt
mhidalgo-bdai: Hmm, do we have to? CMake already bundles and selects [per-compiler configuration](https://github.com/Kitware/CMake/tree/d6e03018cb67698366af0742cfaa49ec8c29d0fb/Modules/Compiler) based on the `CMAKE_BUILD_TYPE` set.

amessing-bdai: You could leave out the `compile_options`, but I still think the `compile_definitions` are useful to have.

amessing-bdai: Still might recommend keeping the `-O0` for debug, but up to you

.pre-commit-config.yaml
mhidalgo-bdai: It was not as easy as I was expecting it to be. Naturally, ROS 2 generated code won't pass `clang-tidy`. `ament_cmake_clang_tidy` wraps `clang-tidy` in enough abstraction layers to make it impossible to narrow down the set of source files it diagnoses.

So, I hooked it up manually in https://github.com/bdaiinstitute/ros_utilities/pull/118/commits/7513af2fa4a8dc7a38ec97060f962c2455c1851d. I also kept `cpplint`. These are not mutually exclusive.

amessing-bdai: They are not, but I've found that cpplint causes a lot more comments to ignore linting than one would like. I don't think its a problem to have it on though.

proto2ros/CMakeLists.txt
mhidalgo-bdai: Done in https://github.com/bdaiinstitute/ros_utilities/pull/118/commits/2c0bf18ebcf5e4cb6951b6cbb02a391b3d442bb1.

amessing-bdai (approved):

mhidalgo-bdai:
Hmm, something's off with generated Protobuf C++ code. It sometimes fails to preprocess some of it, but only in CI, I cannot reproduce it anywhere else :thinking:

PR #123, opened 2024-10-02
Allow adapted feeds to filter messages

mhidalgo-bdai:
This patch adds a simple, generic mechanism for `AdaptedFeed` instances to drop messages.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
@@ -1,12 +1,30 @@
 # Copyright (c) 2024 Boston Dynamics AI Institute Inc.  All rights reserved.
 
-from typing import Any, Callable, Generator, Generic, Iterable, List, Literal, Optional, TypeVar, Union, overload
+from typing import (
+    Any,
+    Callable,
+    Generator,
+    Generic,
+    Iterable,
+    List,
+    Literal,
+    Optional,
+    TypeVar,
+    Union,
+    overload,
+)
 
 import tf2_ros
 from rclpy.node import Node
 
 import bdai_ros2_wrappers.scope as scope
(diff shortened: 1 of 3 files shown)


bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
```diff
@@ -1,12 +1,30 @@
...
+    Generator,
+    Generic,
+    Iterable,
+    List,
+    Literal,
+    Optional,
+    TypeVar,
+    Union,
+    overload,
+)
```
mhidalgo-bdai, line 15: `black` decided it was time to reformat :man_shrugging:

bdai_ros2_wrappers/bdai_ros2_wrappers/filters.py
```diff
@@ -215,15 +222,19 @@ def __init__(self, upstream: Filter, fn: Callable) -> None:
...
+            If none is returned, no message signaling will occur.
         """
         super().__init__()
         self.fn = fn
         self.connection = upstream.registerCallback(self.add)
 
     def add(self, *messages: Any) -> None:
         """Adds new `messages` to the adapter."""
-        self.signalMessage(self.fn(*messages))
+        result = self.fn(*messages)
```
mhidalgo-bdai, line 235: @amessing-bdai ended up going with the trivial implementation. There is a way, an `async` way for `fn` to nicely embed complex logic in a functional way, but it was more trouble than it was worth. If one needs lots of bookkeeping and control flow for an adapter, it may be about time to write a new filter.

bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
amessing-bdai: There is a line length limit for black.

amessing-bdai (approved):

PR #124, opened 2024-10-07
Allow `proto2ros_generate()` to depend on `protobuf_generate()`

mhidalgo-bdai:
Precisely what the title says. Also fixes a race in `proto2ros_tests` as a result of failing to establish this dependency.


proto2ros_tests/CMakeLists.txt
```diff
@@ -45,9 +45,7 @@ proto2ros_generate(
   CPP_OUT_VAR cpp_sources
   INCLUDE_OUT_VAR cpp_include_dir
   APPEND_PYTHONPATH "${PROTO_OUT_DIR}"
-)
-add_dependencies(
-  ${PROJECT_NAME}_messages_gen ${PROJECT_NAME}_proto_gen
```
mhidalgo-bdai, line 50: This adds the dependency to the custom target that is downstream of the actual `proto2ros generate` command, so it doesn't really work.

PR #126, opened 2024-10-17
Improve service and action exception messages

mhidalgo-bdai:
Precisely what the title says. Service and action exceptions could render a more informative message when stringified.


jbarry-bdai (approved):
thank you!

PR #129, opened 2024-11-14
Add support for interruptible ROS processes

mhidalgo-bdai:
By default, `rclpy` will install signal handlers for SIGINT and SIGTERM upon global context initialization. These handlers shutdown ROS 2 contexts. process-wide, making it impossible for ROS 2 aware processes to use most ROS 2 interfaces during  shutdown. E.g. a simple example script cannot command a robot to adopt a safe configuration on Ctrl + C,

This patch adds the notion of an interruptible ROS process. An interruptible ROS process will not shutdown when interrupted (SIGINT or SIGTERM) but raise a suitable exception (KeyboardInterrupt or SystemExit). When using process-wide APIs, the global context will eventually be shutdown accordingly. This is an opt-in feature.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/executors.py
@@ -14,6 +14,7 @@ import typing
 import weakref
 
 import rclpy.executors
+import rclpy.node
 
 from bdai_ros2_wrappers.futures import FutureLike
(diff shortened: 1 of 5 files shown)


khughes-bdai (approved):
LGTM -- eventually would be nice to have an example (maybe located internally?) that uses this

tcappellari-bdai (approved):

mhidalgo-bdai:
With green CI, getting this one in.

PR #132, opened 2024-11-20
Deferred message filter/feed activation for race-free setups

mhidalgo-rai:
Message filters (and feeds too, by transitivity) can easily lose messages when these are chained while messages are being dispatched. This is a very real possibility when e.g. setting up topic subscriptions against a `Node` that is spinning in the background.

To alleviate this, this patch introduces `start()`/`stop()` semantics to both message filters and feeds. Default behavior remains the same, but users can now choose not to `autostart` them and defer it until after all chaining is complete.

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
@@ -197,6 +197,7 @@ class AdaptedMessageFeed(MessageFeed[MessageT]):
         self,
         feed: MessageFeed,
         fn: Callable[..., MessageT],
+        *,
         autostart: bool = True,
         **kwargs: Any,
     ) -> None:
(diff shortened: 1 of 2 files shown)


khughes-bdai (approved):

tcappellari-bdai (approved):
lgtm

bdai_ros2_wrappers/bdai_ros2_wrappers/feeds.py
```diff
@@ -190,6 +197,7 @@ def __init__(
         self,
         feed: MessageFeed,
         fn: Callable[..., MessageT],
+        autostart: bool = True,
```
amessing-bdai, line 200:
```suggestion
        *,
        autostart: bool = True,
```

Make the use of the code more readable

```
AdaptMessageFeed(..., autostart=False) vs AdaptMessageFeed(..., False)
```

amessing-bdai (approved):

PR #136, opened 2024-12-20
[SW-1624] Deprecate old wrapper name

tcappellari-bdai:
## Proposed changes

Deprecate the old name `bdai_ros2_wrappers` for new name `synchros2`

`test_import.py` still passes and prints expected deprecation message

### Checklist

- [ ] Lint and unit tests pass locally
- [ ] I have added tests that prove my changes are effective
- [ ] I have added necessary documentation to communicate the changes

### DIFF
--- a/bdai_ros2_wrappers/bdai_ros2_wrappers/__init__.py
+++ b/bdai_ros2_wrappers/bdai_ros2_wrappers/__init__.py
@@ -3,14 +3,16 @@
 import importlib
 import pkgutil
 import sys
+import warnings
 
-from bdai.deprecation import deprecated
-
-
-@deprecated(
-    reason="bdai_ros2_wrappers has been renamed to synchros2. Please use the renamed version",
-    alternative="synchros2",
+warnings.simplefilter("default")
+warnings.warn(
+    "bdai_ros2_wrappers has been renamed to synchros2. Please use the new name synchros2 instead",
+    DeprecationWarning,
+    stacklevel=2,
 )
+
+
 def aliased_import(name, alias):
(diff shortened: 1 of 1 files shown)


bdai_ros2_wrappers/bdai_ros2_wrappers/__init__.py
```diff
@@ -4,7 +4,13 @@
 import pkgutil
 import sys
 
+from bdai.deprecation import deprecated
```
khughes-bdai, line 7: think we can't use this import here since ros_utilities can't depend on bdai

tcappellari-bdai: ya ive just realized that but the built in version isnt printing properly when i run a test
changing this to a draft in the meantime

jcarpinelli-bdai (approved):

tcappellari-bdai:
> All looks good to me. Just so I understand, what is the behavior that this PR changes? This basically allows someone to type `import bdai_ros2_wrappers` and get a warning telling them to `import synchros2` instead?
> 
> If that's true, and nothing is really (substantive) is really imported with `import bdai_ros2_wrappers`, I wonder if this could / should be an exception or error instead. Something like...
> 
> ```python
> raise ModuleNotFoundError(
> """
> ...renamed to synchros2...
> """)
> ```

I don't want to break anything thats already depending on `bdai_ros2_wrappers` just yet. Currently, importing it will actually import `synchros2` so the warning is a reminder to change it. We'll officially deprecate and put in the module not found error later

PR #140, opened 2025-02-05
Add publisher extension

amessing-bdai:
## Proposed changes

- Added a publisher wrapper type class that contains functions for waiting on subscriptions to connect
- Added unit test

### Checklist

- [X] Lint and unit tests pass locally
- [X] I have added tests that prove my changes are effective
- [X] I have added necessary documentation to communicate the changes

### DIFF
--- a/synchros2/synchros2/publisher.py
+++ b/synchros2/synchros2/publisher.py
@@ -79,3 +79,6 @@ class Publisher(Generic[_MessageT]):
     def publisher(self) -> rclpy.publisher.Publisher:
         """Returns the internal ROS 2 publisher"""
         return self._publisher
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._publisher, name)
--- a/synchros2/synchros2/publisher.py
+++ b/synchros2/synchros2/publisher.py
@@ -1,0 +1,80 @@
+#  Copyright (c) 2025 Boston Dynamics AI Institute Inc.  All rights reserved.
+
+from asyncio import Future
+from typing import Any, Generic, Optional, Type, TypeVar
+
+import rclpy.publisher
+from rclpy.node import Node
+
+import synchros2.scope as scope
+
+# Represents a ros message type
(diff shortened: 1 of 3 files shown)


synchros2/synchros2/publisher.py
```diff
@@ -0,0 +1,81 @@
...
+import rclpy.publisher
+from rclpy.node import Node
+
+import synchros2.scope as scope
+
+# Represents a ros message type
+_MessageT = TypeVar("_MessageT")
+
+
+class Publisher(Generic[_MessageT]):
```
mhidalgo-bdai, line 15: @amessing-bdai meta: a `publish` method would be nice, it could allow in-place message construction for ease of use.

synchros2/synchros2/publisher.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Args:
+            num_subscriptions: lower bound on the number of subscriptions to match.
+
+        Returns:
+            a future, done if the current number of subscriptions already matches
+            the specified lower bound.
+        """
+        future_match = Future()  # type: ignore[var-annotated]
+        num_matched_publishers = self._node.count_subscribers(self._publisher.topic_name)
```
mhidalgo-bdai, line 47: @amessing-bdai nit: would be good to expose the `topic_name` as a property too.

synchros2/synchros2/publisher.py
amessing-bdai: Fair. I was trying to figure out if there was a way to have function call pass through to the internal publisher if they are not defined here. I could inherit from the rclpy Publisher, but a wrapper seems like it will be more extendable in the future.

synchros2/synchros2/publisher.py
amessing-bdai: It should get one from exposing the attr of the publisher

mhidalgo-bdai (approved):

PR #142, opened 2025-02-27
Fix copyrights

baxelrod-bdai:
## Proposed changes

Simply standardizes the copyright text.  This is especially important in public repositories.  This will also make it easier to use common tooling for copyrights if/when further changes need to be made.


mhidalgo-bdai (commented):
Thanks for the patch @baxelrod-bdai! Should copyright be attributed to RAI now?

baxelrod-bdai:
>  Should copyright be attributed to RAI now?

No.  Not yet according to Legal.  But I am about to check in a tool that we can use in pre-commit to handle copyright updates automatically.

mhidalgo-bdai (approved):
LGTM with DCO signature.

PR #152, opened 2025-04-11
[EK-68] Mitigate hangs induced by multi-threaded executors on scope exit

mhidalgo-bdai:
## Proposed changes

Precisely what the title says. When leaving a `ROSAwareScope` context, the underlying executor will attempt a shutdown. For multi-threaded executors, there may be tasks blocked on arbitrary conditions that if never fulfilled can induce a hang. This patch mitigates this issue by shutting down the context associated with the scope, waking up tasks that are context aware (as they should be).

### Checklist

### Additional comments

This is not a general solution. It still relies on a sane codebase. If some callback goes on to block indefinitely, there's nothing this can do to stop it (technically _there is_ but alternatives to this come with plenty of caveats). Proper use of timeouts and context-aware waits (e.g. `synchros2.futures.wait_for_future()`) are necessary.

### DIFF
--- a/synchros2/synchros2/scope.py
+++ b/synchros2/synchros2/scope.py
@@ -164,11 +164,11 @@ class ROSAwareScope(typing.ContextManager["ROSAwareScope"]):
             if self._tf_listener is not None:
                 self._tf_listener.shutdown()
                 self._tf_listener = None
-            assert self._executor is not None
-            for node in self._graph:
-                self._executor.remove_node(node)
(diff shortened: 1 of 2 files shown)


amessing-bdai (approved):

jbarry-bdai (approved):
Yay!  Can we make sure we get the bdai hash updated as soon as this goes in cuz this will be super helpful :D

khughes-bdai (approved):

PR #162, opened 2025-09-30
Getting started guide

jbarry-bdai:
## Proposed changes

Adding a getting started guide that takes users through the most basic use of synchros2 and the associated code.

I intend to add more tutorials so this is set up for future additions.

I would appreciate someone else running the tutorial.  I have tried it locally.

### Checklist

### DIFF
--- a/synchros2/docs/GettingStartedGuide.md
+++ b/synchros2/docs/GettingStartedGuide.md
@@ -1,0 +1,236 @@
+# Getting Started Guide
+
+# Installation
+
+## Prerequisites
+
+These links are to Humble tutorials but `synchros2` should be compatible with any current ROS2 distro.
+
+1. [Configure a ROS2 environment](https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Configuring-ROS2-Environment.html)
+2. [Create a ROS2 workspace](https://docs.ros.org/en/humble/Tutorials/Beginner-Client-Libraries/Colcon-Tutorial.html)
+
+## Install and Build `synchros2`
+
+1. Download the git repository into your existing ROS workspace:
+    
+    ```bash
+    cd <workspace>/src
+    git clone https://github.com/bdaiinstitute/ros_utilities.git
+    ```
+    
+2. Build and source the colcon workspace
+    
+    ```bash
+    cd <workspace>
+    colcon build --symlink-install
+    source install/setup.bash
+    ```
+    
+
+# Overview
+
+`synchros2` is a framework that allows blocking in callbacks without the user needing to worry about the threading.  The simplest way to use it is to:
+
+1. Decorate your `main` with the `@ros_process.main()` decorator:
+    
+    ```python
+    import synchros2.process as ros_process
+    
+    @ros_process.main()
+    def main() -> None:
+    	...
+    ```
+    
+2. Only use the node associated with `ros_scope`:
+    
+    ```python
+    import synchros2.scope as ros_scope
+    
+    class MyClass:
+    	def __init__(self):
+    		self._node = ros_scope.ensure_node()
+    ```
+    
+
+# Your First `synchros2` Code
+
+We’ll now create and run example `synchros2` code.  We’ll analyze the code in the next section.
+
+1. Create a package called `synchros2_tutorials` that depends on `synchros2` (by specifying `--dependencies synchros2` the package creation will automatically add a dependency on `synchros2` to the created `package.xml`): 
+    
+    ```bash
+    cd <workspace>/src
+    ros2 pkg create --build-type ament_python --license Apache-2.0 synchros2_tutorials  --dependencies synchros2
+    ```
+    
+2. Create the file `<workspace>/src/synchros2_tutorials/synchros2_tutorials/listener.py` and add the following code to it:
+    
+    ```python
+    import std_msgs.msg
+    import time
+    
+    import synchros2.process as ros_process
+    import synchros2.scope as ros_scope
+    
+    class Listener:
+        def __init__(self):
+            self._node = ros_scope.ensure_node()
+            self._sub1 = self._node.create_subscription(std_msgs.msg.String, "chat", self._callback1, 1)
+            self._sub2 = self._node.create_subscription(std_msgs.msg.String, "chat", self._callback2, 1)
+            self._node.get_logger().info("Listening!")
+    
+        def _callback1(self, msg: std_msgs.msg.String) -> None:
+            self._node.get_logger().info(f"Callback 1 received message {msg} and will now sleep for 10 seconds")
+            time.sleep(10)
+            self._node.get_logger().info(f"Callback 1 is done sleeping after receiving {msg}")
+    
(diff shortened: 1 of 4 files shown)


synchros2/docs/GettingStartedGuide.md
```diff
@@ -0,0 +1,236 @@
...
+    
+        def _callback2(self, msg: std_msgs.msg.String) -> None:
+            self._node.get_logger().info(f"Callback 2 received message {msg} and will now sleep for 5 seconds")
+            time.sleep(5)
+            self._node.get_logger().info(f"Callback 2 is done sleeping after receiving {msg}")
+    
+    @ros_process.main()
+    def main() -> None:
+        _ = Listener()
+        ros_process.wait_for_shutdown()
```
mhidalgo-bdai, line 95: @jbarry-bdai meta: looking at this laid out like that, I feel like we need a better pattern than this construct-and-forget thing (though it is what it is and we would mostly be hiding it). Maybe something like:

```
@ros_process.main(cli())
def main(args):
     ros_process.exec(Listener(args))
```
and then have Listener optionally define a `__call__` override. If it does, it'll get called, if it doesn't, exec will just wait until shutdown.

synchros2/docs/GettingStartedGuide.md
```diff
@@ -0,0 +1,236 @@
...
+@ros_process.main()
+def main() -> None:
+```
+
+**This decorator is very important!**  This is what creates the ros process node and the corresponding executor that keeps track of all of the threads.
+
+Inside the main we create an instance of the listener class and then we wait for shutdown.  This leaves our subscriptions running until the program receives `Ctrl+C`:
+
+```python
+    _ = Listener()x
```
mhidalgo-bdai, line 191: @jbarry-bdai extra x:


```suggestion
    _ = Listener()
```

synchros2/docs/GettingStartedGuide.md
tcappellari-bdai: I agree with jenny that its tidier but a lot more confusing for a new user

tcappellari-bdai (approved):
lgtm

synchros2/docs/GettingStartedGuide.md
mhidalgo-bdai: Fair. I'm not very fond of the constructor-does-things pattern, but I guess folks can implement their own variations of `Listener.run()` if they feel like it. Maybe that's an interesting addition: a tutorial where the process does things at a fixed rate in the main thread (ROS things or not).

synchros2/docs/GettingStartedGuide.md
```diff
@@ -0,0 +1,236 @@
...
+We’ll now create and run example `synchros2` code.  We’ll analyze the code in the next section.
+
+1. Create a package called `synchros2_tutorials` that depends on `synchros2` (by specifying `--dependencies synchros2` the package creation will automatically add a dependency on `synchros2` to the created `package.xml`): 
+    
+    ```bash
+    cd <workspace>/src
+    ros2 pkg create --build-type ament_python --license Apache-2.0 synchros2_tutorials  --dependencies synchros2
+    ```
+    
+2. Create the file `<workspace>/src/synchros2_tutorials/synchros2_tutorials/listener.py` and add the following code to it.
```
mhidalgo-bdai, line 66: @jbarry-bdai nit: 
```suggestion
2. Create the file `<workspace>/src/synchros2_tutorials/synchros2_tutorials/listener.py` and add the following code to it:
```

synchros2/docs/GettingStartedGuide.md
```diff
@@ -0,0 +1,236 @@
...
+
+We then create a class that stores the ros scope node:
+
+```python
+class Listener:
+    def __init__(self):
+        self._node = ros_scope.ensure_node()
+```
+
+The `synchros` mechanisms create a single node that we want to use everywhere because it's already set up to create all the appropriate threads for any callbacks added to it.  We refer to this single node as the "ros scope node".  For beginning use, the ros scope node is the *only* node you need or should use.  The `synchros2` mechanics take care of spinning any callbacks that we attach to this node and spawning whatever threads we need.  This node can be obtained anywhere in the process with the `ros_scope.ensure_node()` call (which will either return the ros scope node or raise a `ValueError` if one was never created - usually because you forgot to decorate your `main` with `ros2_process.main()`).  **Do not create further nodes.**
```
mhidalgo-bdai, line 153: @jbarry-bdai nit: 
```suggestion
The `synchros2` mechanisms create a single node that we want to use everywhere because it's already set up to create all the appropriate threads for any callbacks added to it.  We refer to this single node as the "ros scope node".  For beginning use, the ros scope node is the *only* node you need or should use.  `synchros2` takes care of spinning any callbacks that we attach to this node and spawning whatever threads we need.  This node can be obtained anywhere in the process with the `ros_scope.ensure_node()` call (which will either return the ros scope node or raise a `ValueError` if one was never created - usually because you forgot to decorate your `main` with `ros2_process.main()`).  **Do not create further nodes.**
```

synchros2/docs/GettingStartedGuide.md
```diff
@@ -0,0 +1,236 @@
...
+4. Immediately after 2: *While callback 1 is running*, callback 2 also received the first message, printed out the info and started to sleep
+5. Slightly later: We published the message “This is the second message” on `/chat`
+6. Immediately after 5: That message went into the queues for both callbacks.  However the callbacks were both busy so nothing happens yet.
+7. 5 seconds after 4: Callback 2 finishes sleeping and prints that it is done sleeping after receiving the first message
+8. Immediately after 7: Callback 2 is given the second message now that it’s finished with the first and prints that it’s got that
+9. 10 seconds after 3: Callback 1 finishes sleeping and prints that it is done sleeping after receiving the first message
+10. Immediately after 9: Callback 1 is given the second message now that it’s finished with the first and prints that it’s got that
+11. 5 seconds after 7: Callback 2 finishes sleeping and prints that it’s done sleeping after the second message
+12. 10 seconds after 10: Callback 1 finishes sleeping and prints that it’s done sleeping after the second message
```
mhidalgo-bdai, line 227: @jbarry-bdai meta: if you toss this into Gemini or GPT, you can get a nice [`mermaid` sequence diagram](https://mermaid.js.org/syntax/sequenceDiagram.html).

synchros2/docs/GettingStartedGuide.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+Our main function is decorated with `ros_process.main()`:
+
+```python
+@ros_process.main()
+def main() -> None:
+```
+
+**This decorator is very important!**  This is what creates the ros process node and the corresponding executor that keeps track of all of the threads.
```
mhidalgo-bdai, line 186: @jbarry-bdai consider: 
```suggestion
**This decorator is very important!**  This is what creates the process-wide ros scope node and the corresponding executor that keeps track of all of the threads.
```
or similar to connect the notions of process node with scope node without explaining much.

mhidalgo-bdai (approved):

PR #164, opened 2025-10-08
Move synchros2 tutorials to their own package and create custom message

jbarry-bdai:
## Proposed changes

This creates `synchros2_tutorials` and `synchros2_tutorials_interfaces` packages as examples in `ros_utilities`.  Since I didn't want people to have package conflict after creating their own versions I named the ones here with `_example`.

Additionally, I updated the tutorials to use a custom string message instead of `std_msgs/String`.  `std_msgs/String` is deprecated and we'll need the interfaces package for the action tutorial.  However it does add a bit of complexity to `GettingStartedGuide` so I'm open to using the example messages string instead.

### Checklist

### DIFF
--- a/synchros2/docs/GettingStartedGuide.md
+++ b/synchros2/docs/GettingStartedGuide.md
@@ -29,7 +29,7 @@ These links are to Humble tutorials but `synchros2` should be compatible with an
 
 ## Overview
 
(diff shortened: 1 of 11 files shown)


mhidalgo-bdai (commented):
First pass.

PR #167, opened 2025-10-15
[SW-3230] Simplify prebaked node naming

mhidalgo-rai:
## Proposed changes

This patch simplifies the whole naming story for the implicit node in `ros_process`.

### Checklist

### DIFF
--- a/synchros2/synchros2/process.py
+++ b/synchros2/synchros2/process.py
@@ -53,9 +53,10 @@ class ROSAwareProcess:
         *,
         uses_tf: bool = False,
         interruptible: bool = False,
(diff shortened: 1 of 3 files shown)


tcappellari-bdai (approved):

jbarry-bdai (commented):
This seems much simpler!

But... it's not backwards compatible.  I suspect there are things to fix in `bdai`.  Are we ok with this change for anyone else who might be using this?  Should we perhaps deprecate letting prebaked be a string and print a warning but not remove that yet?

jbarry-bdai:
Let's be rebels and not worry about backwards compatibility.  Good to get the APIs sorted before we upstream.

PR #169, opened 2025-10-27
Fix mypy errors

jbarry-bdai:
## Proposed changes

These were causing mypy errors (even though I hadn't actually touched these files...)

### Checklist

### DIFF
--- a/.configs/mypy.ini
+++ b/.configs/mypy.ini
@@ -1,5 +1,5 @@
 [mypy]
-mypy_path = $MYPY_CONFIG_FILE_DIR/..
+mypy_path = $MYPY_CONFIG_FILE_DIR/../synchros2
 python_version = 3.8
 disallow_untyped_defs = True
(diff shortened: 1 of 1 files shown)


mhidalgo-bdai (approved):
Do we still need this patch @jbarry-bdai ?

jbarry-bdai:
> Do we still need this patch @jbarry-bdai ?

I still see them on main.  In fact another one has crept up!  Not sure how this happens :-p  I'll fix the new one too

image

jbarry-bdai:
@mhidalgo-bdai it happens in the docker container but not outside of it.  pre-commit version inside docker is 3.5.0, outside 4.20.  not sure what else would matter.  they're legitimate issues though?

mhidalgo-bdai:
> they're legitimate issues though?

@jbarry-bdai it looks like it, yes, but if we cannot reproduce it's hard to fix and track them.

PR #189, opened 2025-11-21
Fix import in executors example

szobov:


szobov:
Hey @mhidalgo-bdai 
Done.

I've done the fix though the github interface and didn't see signoff is needed.

PR #203, opened 2026-04-10
Improve consistency of timers and rates

mhidalgo-rai:
## Proposed changes

This patch does two things:

- It adds an optional dedicated thread pool for timers only to our executor (default enabled) 
- It adds an `rclpy.Rate` re-implementation that uses clocks directly rather timers

### Checklist

### Additional comments

Initial testing on a simulated stack shows promise.

### DIFF
--- a/synchros2/synchros2/node.py
+++ b/synchros2/synchros2/node.py
@@ -4,12 +4,14 @@ import functools
 from typing import Any, Callable, Iterable, Optional, Type
 
 from rclpy.callback_groups import CallbackGroup
+from rclpy.clock import Clock
 from rclpy.exceptions import InvalidHandle
 from rclpy.node import Node as BaseNode
 from rclpy.waitable import Waitable
 
 from synchros2.callback_groups import NonReentrantCallbackGroup
 from synchros2.logging import MemoizingRcutilsLogger
+from synchros2.time import SteadyRate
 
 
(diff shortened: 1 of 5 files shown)


synchros2/synchros2/node.py
```diff
@@ -55,6 +57,28 @@ def default_callback_group(self) -> CallbackGroup:
         # NOTE(hidmic): this overrides the hardcoded default group in rclpy.node.Node implementation
         return self._default_callback_group_override
 
+    def create_rate(
```
jbarry-bdai, line 68: should this be marked override?

jbarry-bdai (approved):

synchros2/synchros2/node.py
mhidalgo-rai: We are technically not overriding but shadowing: return types are different. Hmm, but maybe we could make it look like `Rate` was returned. Let me see.

mhidalgo-rai: PTAL @ https://github.com/bdaiinstitute/synchros2/pull/203/commits/230c70270d514e1db6c16c896ef1b233c8164689

mhidalgo-rai:
@tcappellari-bdai mind to take a look before merging?

mhidalgo-rai:
Alright, moving forward.

PR #204, opened 2026-04-14
Handle missing `typing.override`

mhidalgo-rai:
## Proposed changes

Follow-up to #203. There is no `typing.override` back in Python 3.10, and the `typing_extensions` version distributed with Ubuntu Jammy is old enough to lack a forward port. This patch just replaces it with an identity decorator.

### Checklist

- [ ] I have added tests that prove my changes are effective
- [ ] I have added necessary documentation to communicate the changes

### DIFF
--- a/synchros2/synchros2/node.py
+++ b/synchros2/synchros2/node.py
@@ -6,8 +6,7 @@ from typing import Any, Callable, Iterable, Optional, Type
 try:
     from typing import override  # type: ignore[attr-defined]
 except ImportError:
-    from typing_extensions import override  # type: ignore[import]
-
+    override = lambda func: func  # noqa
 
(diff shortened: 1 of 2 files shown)


jbarry-bdai (approved):
does `from overrides import override` work?  that's what i've always used

taughz (approved):

mhidalgo-rai:
> does from `overrides` import `override` work? that's what i've always used

That one is [another package](https://pypi.org/project/overrides/), and it's [not distributed](https://github.com/ros/rosdistro/blob/0fee186331c1ba262af02cc6826f1e903fd1fbe4/rosdep/python.yaml#L8172-L8185) with Ubuntu, only through PyPI, so we can't depend on it.

